Issue #18, opened 2014-12-17
Snap planner is NOT collision checking and docstring does NOT specify that


siddhss5:
There's absolutely no wording in the docstring that screams out that the snap planner is not checking collisions. It's merely creating a trajectory with two points and passing it merrily on. This is very dangerous if misused.
https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/snap.py#L43

Issue #17, opened 2014-12-17
Snap tolerance breaks if DOFs have mixed units [revolute + prismatic]


siddhss5:
No reason why DOFs should exclusively be in radians.
https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/snap.py#L48

Wouldn't it be better if we just provided a function pointer that evaluates to true when close enough?

Issue #16, opened 2014-12-17
Snap planner is not checking absolute value of difference for tolerance check


siddhss5:
Here's the offending line:
https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/snap.py#L93

PR #19, opened 2014-12-17
Bugfixes for SnapPlanner

mkoval:
I made the following changes:
1. Added the missing absolute value in our distance calculation (#16)
2. Removed the `snap_tolerance` parameter and, instead, use the DOF resolutions (#17)
3. Check collision at the start and end-points of the trajectory (#18)

I am holding off on merging the pull request until I do some more testing. It would also be good to get feedback from @siddhss5.

### DIFF
--- a/tests/test_SnapPlanner.py
+++ b/tests/test_SnapPlanner.py
@@ -143,6 +143,17 @@ class SnapPlannerTest(unittest.TestCase):
         with self.assertRaises(PlanningError):
             self.planner.PlanToConfiguration(self.robot, self.config_segment2)
 
+    def test_CloneRobot(self):
+        self.robot.SetActiveDOFValues(self.config_self_collision)
+        self.assertTrue(self.robot.CheckSelfCollision())
+
+        cloned_robot = openravepy.RaveCreateRobot(self.env, self.robot.GetXMLId())
+        cloned_robot.Clone(self.robot, 0)
+        cloned_robot.SetName(self.robot.GetName() + '_clone')
+        self.env.Add(cloned_robot)
+
+        self.assertTrue(cloned_robot.CheckSelfCollision())
+
     def test_CloneBroken(self):
         self.robot.SetActiveDOFValues(self.config_self_collision)
         self.assertTrue(self.robot.CheckSelfCollision())
@@ -154,7 +165,7 @@ class SnapPlannerTest(unittest.TestCase):
         self.assertTrue(cloned_robot.CheckSelfCollision())
 
     def test_CloneWorking(self):
-        self.robot.SetActiveDOFValues(self.config_env_collision)
+        self.robot.SetActiveDOFValues(self.config_self_collision + 10 * self.config_epsilon)
 
         cloned_env = openravepy.Environment()
         cloned_env.Clone(self.env, openravepy.CloningOptions.Bodies)


mkoval:
I switched to collision checking the straight-line trajectory, instead of only allowing short snaps. Instead of implementing the collision detection manually, I call the built-in `CheckPathAllConstraints` function. This is the same function that built-in OpenRAVE planners (e.g. BiRRT) use to collision check edges. I am not sure whether that function does a bisection test or not.

I also added a few unit tests. [The test](https://github.com/personalrobotics/prpy/blob/bugfix/SnapPlanner/tests/test_SnapPlanner.py#L120) that checks self-collisions fails. This is quite surprising, because [a nearly identical test](https://github.com/personalrobotics/prpy/blob/bugfix/SnapPlanner/tests/test_SnapPlanner.py#L108) that checks environment collisions passes.

I haven't been able to figure out why. The logic inside `CheckPathAllConstraints` is incredibly complicated:

1. `CheckPathAllConstraints` [calls `_checkpathvelocityconstraintsfn`](https://github.com/rdiankov/openrave/blob/7d0dc9c945bab4fc6e3a703f707bb64656b7b897/include/openrave/planner.h#L238) on the `PlannerParameters` struct
2. The filter options are set to `0xFFFF` by default, which includes: ([see here](https://github.com/rdiankov/openrave/blob/7d0dc9c945bab4fc6e3a703f707bb64656b7b897/include/openrave/planner.h#L30))
    * `CFO_CheckEnvCollisions`
    * `CFO_CheckSelfCollisions`
    * `CFO_CheckTimeBasedConstraints`
    * `CFO_BreakOnFirstValidation`
2. `_checkpathvelocityconstraintsfn` [defaults to `DynamicsCollisionConstraint::Check`](https://github.com/rdiankov/openrave/blob/8e2ac09a4c91e63c34c0a01e675ebe38f78de6d0/src/libopenrave/planner.cpp#L441)
3. [`DynamicsCollisionConstraint::Check`](https://github.com/rdiankov/openrave/blob/2f7c8a6e582fd74746902e63b5707052fee87b09/src/libopenrave/planningutils.cpp#L2122) checks both geometric and dynamic feasibility
    * Passing empty vectors for `dq0` and `dq1` seems to disable checks for dynamic feasibility ([this case](https://github.com/rdiankov/openrave/blob/2f7c8a6e582fd74746902e63b5707052fee87b09/src/libopenrave/planningutils.cpp#L2339))
    * There is a suspicious comment that says "this does not check the end config, and may or may not check the start based on the value of 'start'"
4. `Check` calls the [`_SetAndCheckState` function](https://github.com/rdiankov/openrave/blob/2f7c8a6e582fd74746902e63b5707052fee87b09/src/libopenrave/planningutils.cpp#L1982), which sets the robot's configuration
5. `_SetAndCheckState` calls the [`_CheckState` function](https://github.com/rdiankov/openrave/blob/2f7c8a6e582fd74746902e63b5707052fee87b09/src/libopenrave/planningutils.cpp#L2020) to actually do the collision checking

I haven't been able to figure out where in this labyrinth is amiss.

@psigen @siddhss5 @cdellin: Any suggestions?

siddhss5:
Yikes! You're not going to like this but I'd suggest we get some 'ground truth' by writing a simple line segment checker on our own, maybe even with just increments.

I was thinking it might be nice to write out the 'checkOrder' function as a generator allowing us explore different orders.

Sidd.

(Rapidly written in motion)

mkoval:
It turns out that `CheckPathAllConstraints` works fine. There is a more fundamental issue here: `CheckSelfCollision` returns `True` in the original environment, but `False` in the cloned environment. Both environments look identical in the viewer, so this seems to be some sort of cloning issue.

mkoval:
It gets even weirder. This test case works: 808b55ad86b09ded4837bfcda66e8d9eefc82494.

mkoval:
I opened an OpenRAVE issue: https://github.com/rdiankov/openrave/issues/335. I'm out of ideas here, so hopefully Rosen can point us in the right direction.

siddhss5:
My guess is that the clone is somehow mangling the adjacency list and skipping some/all link pair collision checks. Does 'verbose' print out enough info to verify this? Or maybe the collision report?

Sidd.

(Rapidly written in motion)

mkoval:
Nope. `CheckSelfCollision` immediately returns false without any logging. It also doesn't fill in the `CollisionReport`.

Rosen thinks this is related to the assumption that initial configuration of the robot (by default: the zero configuration) is not in self collision. This is used to initialize the adjacency list when loading the robot. I had the same idea, but `GetAdjacentLinks()` returns the same values before and after cloning.

PR #29, opened 2015-01-05
Expose OpenRAVE's builtin planners as prpy Planners.

siddhss5:
I wrote a little wrapper class that exposes OpenRAVE's builtin planners as prpy Planners. I was careful to remove Rosen's arbitrary defaults. I also added support for planning to multiple configurations with OpenRAVE's birrt planner.

### DIFF
--- a/src/prpy/planning/openrave.py
+++ b/src/prpy/planning/openrave.py
@@ -1,0 +1,95 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2013, Carnegie Mellon University
+# All rights reserved.
+# Authors: Siddhartha Srinivasa <siddh@cs.cmu.edu>
+# Authors: Michael Koval <mkoval@cs.cmu.edu>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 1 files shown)


psigen:
Looks good.  Should we do any sort of error checking before `.ravel()` in PlanToConfigurations?

Otherwise if we provide the wrong size goals, openrave might start planning to a shorter or longer list of entirely invalid goals.

siddhss5:
There's also much code duplication with ompl.py. I wonder if `OMPLPlanner` should inherit from `OpenRAVEPlanner` instead of `BasePlanner`:
https://github.com/personalrobotics/prpy/blob/feature/OpenRAVEPlanners/src/prpy/planning/ompl.py#L34

siddhss5:
I added some error checking around `.ravel()` in the latest commit:
https://github.com/personalrobotics/prpy/commit/f8276ad53f789b80eb4cbc73e8b5788f913341a2

It doesn't catch *everything* but at the very least will fail and not plan to a wrong configuration. It'd be nice to figure out a more central place to put some of this.

psigen:
I added a few lint fixes and added the BiRRT planner to prpy.planner.__init__.py.

PR #38, opened 2015-01-26
Changed metaplanners to catch PlanningErrors instead of all Exceptions.

psigen:
This allows client code to actually catch and inspect unexpected errors thrown
from planners, at the expense of being less robust to strange error conditions.
Since these error conditions should be identified and fixed or handled anyway,
this seems more desirable than the previous blanket exception catching strategy.


mkoval:
I'd like to merge this ASAP because it really does help with debugging. However, like @psigen said, we are likely throwing some exceptions of incorrect types in our existing planners.

@jeking04 Are you okay with me merging this now? Or should I wait until after the demo?

PR #42, opened 2015-02-04
Feature: Planning futures

psigen:
This PR adds the ability to pass a `defer=True` flag and an optional `executor` to planning methods.  The `defer` flag causes planning operations to be submitted to a thread executor and immediately return a `Future` which can be queried for the result of the planning call.  This can also be used with `execute=True` to execute trajectories concurrently with other operations.

Use of the `defer` flag requires the installation of python-trollius (https://bitbucket.org/enovance/trollius), a backport of the Python 3 asyncio library.  This dependency is not required if the `defer` flag is not used.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -293,6 +293,13 @@ class Robot(openravepy.Robot):
 
         # Optionally execute the trajectory.
         if 'execute' not in kw_args or kw_args['execute']:
-            return self.ExecuteTrajectory(traj, **kw_args)
+            if 'defer' in kw_args and kw_args['defer'] is True:
+                import trollius
+                with kw_args['executor'] or \
+                        trollius.executor.get_default_executor() as executor:
+                    return executor.submit(self.ExecuteTrajectory,
+                                           traj.result(), kw_args)
+            else:
+                return self.ExecuteTrajectory(traj, **kw_args)
         else:
             return traj
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -284,9 +284,6 @@ class Robot(openravepy.Robot):
         return False
 
     def _PlanWrapper(self, planning_method, args, kw_args):
-        print "H"
-        import IPython
-        IPython.embed()
 
         # Call the planner.
         traj = planning_method(self, *args, **kw_args)
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -307,7 +307,7 @@ class Robot(openravepy.Robot):
         def defer_trajectory(traj_future, kw_args):
             postprocess_trajectory(traj_future.result(), kw_args)
 
-        # Return the trajectory result or a future to the result.
+        # Return either the trajectory result or a future to the result.
         if 'defer' in kw_args and kw_args['defer'] is True:
             from trollius.executor import get_default_executor
             with kw_args.get('executor') or get_default_executor() as executor:
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -312,6 +312,7 @@ class Robot(openravepy.Robot):
 
             # Optionally execute the trajectory.
             if 'execute' not in kw_args or kw_args['execute']:
+                kw_args['defer'] = False
                 return self.ExecuteTrajectory(traj, **kw_args)
             else:
                 return traj
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -305,7 +305,7 @@ class Robot(openravepy.Robot):
 
         # Perform postprocessing on a future trajectory.
         def defer_trajectory(traj_future, kw_args):
-            postprocess_trajectory(traj_future.result(), kw_args)
+            return postprocess_trajectory(traj_future.result(), kw_args)
 
         # Return either the trajectory result or a future to the result.
         if 'defer' in kw_args and kw_args['defer'] is True:
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -284,6 +284,10 @@ class Robot(openravepy.Robot):
         return False
 
     def _PlanWrapper(self, planning_method, args, kw_args):
+        print "H"
+        import IPython
+        IPython.embed()
+
         # Call the planner.
         traj = planning_method(self, *args, **kw_args)
 
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -183,14 +183,27 @@ class Robot(openravepy.Robot):
         raise PrPyException("Path retimer failed with status '{:s}'"
                             .format(status))
 
-    def ExecuteTrajectory(self, traj, retime=True, timeout=None, **kw_args):
+    def ExecuteTrajectory(self, traj, retime=True, timeout=None,
+                          defer=False, executor=None, **kw_args):
         """
         Executes a trajectory and optionally waits for it to finish.
+
+        Passing `defer=True` to this function submits it for background
+        execution, and returns a Future which contains the result.
+
         @param traj input trajectory
         @param retime optionally retime the trajectory before executing it
         @param timeout duration to wait for execution
-        @returns final executed trajectory
+        @param defer return a future to this function and run in the background
+        @returns final executed trajectory or a Future to this result
         """
+        # If the planning call is deferred, submit the call to the executor.
+        if defer is True:
+            from trollius.executor import get_default_executor
+            executor = executor or get_default_executor()
+            return executor.submit(self.ExecuteTrajectory, traj, retime,
+                                   timeout, defer=False, **kw_args)
+
         # Check if this is a base trajectory.
         has_base = hasattr(self, 'base')
         needs_base = util.HasAffineDOFs(traj.GetConfigurationSpecification())
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -310,7 +310,7 @@ class Robot(openravepy.Robot):
         # Return either the trajectory result or a future to the result.
         if 'defer' in kw_args and kw_args['defer'] is True:
             from trollius.executor import get_default_executor
-            with kw_args.get('executor') or get_default_executor() as executor:
-                return executor.submit(defer_trajectory, result, kw_args)
+            executor = kw_args.get('executor') or get_default_executor()
+            return executor.submit(defer_trajectory, result, kw_args)
         else:
             return postprocess_trajectory(result, kw_args)
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -286,21 +286,32 @@ class Robot(openravepy.Robot):
     def _PlanWrapper(self, planning_method, args, kw_args):
 
         # Call the planner.
+        config_spec = self.GetActiveConfigurationSpecification()
         traj = planning_method(self, *args, **kw_args)
 
-        # Strip inactive DOFs from the trajectory.
-        config_spec = self.GetActiveConfigurationSpecification()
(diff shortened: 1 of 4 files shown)


src/prpy/base/robot.py
```diff
@@ -284,15 +284,33 @@ def ViolatesVelocityLimits(self, traj):
...
+            # Strip inactive DOFs from the trajectory.
+            openravepy.planningutils.ConvertTrajectorySpecification(
+                traj, config_spec
+            )
+
+            # Optionally execute the trajectory.
+            if 'execute' not in kw_args or kw_args['execute']:
+                return self.ExecuteTrajectory(traj, **kw_args)
+            else:
+                return traj
```
mkoval: Does this function return a trajectory or a future trajectory? My understanding is that `traj` is a trajectory and `ExecuteTrajectory` returns a deferred trajectory. If that is the case, then this is a bit weird.

src/prpy/planning/base.py
```diff
@@ -52,13 +52,27 @@ def __init__(self, func):
...
-        with Clone(env, clone_env=instance.env):
-            planning_traj = self.func(instance, Cloned(robot), *args, **kw_args)
-            traj = openravepy.RaveCreateTrajectory(env, planning_traj.GetXMLId())
-            traj.Clone(planning_traj, 0)
-
-        return traj
+        defer = kw_args.get('defer')
+
+        with Clone(env, clone_env=instance.env, lock=True, unlock=False) \
+                as cloned_env:
```
mkoval: This is one case where PEP-8 is just plain wrong.

src/prpy/planning/base.py
```diff
@@ -52,13 +52,27 @@ def __init__(self, func):
...
+
+            def call_planner():
+                planning_traj = self.func(instance, Cloned(robot),
```
mkoval: This looks like a potential race condition. `Cloned()` is only valid while in the `with` block. Since `call_planner` is called by the executor, it may not be executed until after we `__exit__` the `with`-block. To be safe, you should resolve the `Cloned()` call outside the `call_planner()` function, but inside the `with`-block.

src/prpy/planning/base.py
```diff
@@ -52,13 +52,27 @@ def __init__(self, func):
...
+                                          *args, **kw_args)
+                traj = openravepy.RaveCreateTrajectory(
+                    env, planning_traj.GetXMLId()
+                )
+                traj.Clone(planning_traj, 0)
+                cloned_env.Unlock()
```
mkoval: The planning environment may not be unlocked if `self.func` raises an exception. The `Unlock()` call needs to be in a `finally`-block. Or, alternative, you can use `contextlib` to manage the `Clone()` context manager.

src/prpy/planning/base.py
```diff
@@ -211,69 +234,50 @@ def get_planners(self, method_name):
...
+            except PlanningError as e:
+                logger.warning("Planning with {:s} failed: {:s}"
+                               .format(planner, e))
+                results[index] = e
+
+        # Call every planners in parallel using a concurrent executor and
+        # return the first non-error result in the ordering when available.
+        from trollius.executor import get_default_executor
+        executor = kw_args.get('executor') or get_default_executor()
+        for _ in executor.map(call_planner, planners):
```
mkoval: What does it mean to iterate over the result of `executor.map`?

psigen: It waits for the next available result and returns it, until all results are computed.

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map

src/prpy/base/robot.py
psigen: It returns a trajectory if defer=False/None, and a future if defer=True, just like all of the other planning methods.

ExecuteTrajectory never returns a deferred trajectory on its own right now.  Should it?

src/prpy/planning/base.py
mkoval: I was looking at that documentation myself. But I still don't see where it defines that behavior. The documentation says:

> Equivalent to `map(func, *iterables)` except `func` is executed asynchronously and several calls to `func` may be made concurrently. The returned iterator raises a `TimeoutError` if `__next__()` is called and the result isn’t available after timeout seconds from the original call to `Executor.map()`. timeout can be an `int` or a `float`. If `timeout` is not specified or `None`, there is no limit to the wait time. If a call raises an exception, then that exception will be raised when its value is retrieved from the iterator.

The documentation for `map(func, *iterables)` says:

> Return an iterator that applies function to every item of `iterable`, yielding the results. If additional iterable arguments are passed, function must take that many arguments and is applied to the items from all iterables in parallel. With multiple iterables, the iterator stops when the shortest iterable is exhausted. For cases where the function inputs are already arranged into argument tuples, see itertools.starmap().

It's not obvious from that text that `map` returns the *next available* result. An implementation that sequentially blocks on each element seems equally valid. Am I missing something here?

src/prpy/base/robot.py
mkoval: I'm not sure.

`ExecuteTrajectory` only returns the trajectory in case it is modified internally (e.g. simplified, smoothed, or re-timed). In my refactor, this requires one or more trips through the planning pipeline and could return a future trajectory.

On the other hand, it would be more useful to return a `future` that tracks the state of the trajectory being executed. This would eliminate the need for `WaitForController` and the tri-state `timeout` named parameter to `ExecuteTrajectory`. It would also, finally, give us a good method of querying error codes from the controller (e.g. stalled, aborted, etc).

Thoughts?

src/prpy/planning/base.py
psigen: Whoops, this is actually supposed to be as_completed()

src/prpy/base/robot.py
psigen: If ExecuteTrajectory throws a TrajectoryExecutionError for error codes, we
could easily set_error() those in the future and return the executed
trajectory as a set_result() on success.

It is useful to return the trajectory as executed in both cases for
replanning and fault recovery purposes.  We should probably include the
executed traj until the fault in the aforementioned
TrajectoryExecutionError.

mkoval: Currently, `ExecuteTrajectory` can raise `TrajectoryExecutionError` exceptions if a `timeout` is specified. This is an unfortunate side-effect of the current blocking implementation.

I'd prefer to make `ExecuteTrajectory` always use a future internally. The effect of passing `defer=False`, if a timeout is specified, is simply to wait on the future before returning. The downside of this approach is that PrPy must depend on Trollius, even if `defer` is never being passed.

psigen: I don't quite follow.  Is there some reason we can't raise that error for
other conditions, and have the timeout be one particular subclass?  That
seems like a normal use case.

I also don't quite understand how using a future internally helps with
anything.  Perhaps I'm missing some additional details.

mkoval: In the current `master`, we don't consider timing out to be an error. The function silently returns without raising an exception. Also, a `TrajectoryExecutionError` can only be raised if the trajectory finishes executing before the timeout expires.

I am fine with changing this. One downside of the change is that the `timeout = 0` case becomes quite awkward. We regularly use `timeout = 0` when we don't want to wait for the controller at all. With this change, the function will immediately raise a `TimeoutException` which will need to be caught by user code.

mkoval: On second thought, `ExecuteTrajectory` should really have two return values (i.e. a tuple):

1. Actual trajectory being executed after any pre-processing (e.g. cleanup, retiming). This could be a future trajectory if we think the pre-processing code will be time consuming.
2. Future trajectory status that is evaluated when the trajectory finishes execution. This includes any error information.

All of our existing `timeout` functionality can be handled by blocking on the future with a timeout.

psigen: I think that timing out should definitely be considered a type of exception.

It seems like `timeout=0` should throw an exception immediately, as it implies requiring the trajectory to complete immediately, and the case of not wanting a timeout at all should be handled by `timeout=None`.

psigen: Sure, this makes sense for the defer=True case.

psigen: I still think we should have the future return the trajectory as it was
executed.  This is immensely useful when doing force moves or move until
touch stuff.

mkoval: We currently have three cases:

1. **Block, No Timeout:** wait forever (currently `timeout=None`, default)
2. **Block, Timeout:** wait for `timeout` seconds, then raise a `TimeoutException` (currently `timeout=5.0`)
3. **No Block:** return immediately (currently `timeout=0`)

I am very hesitant to switch the meaning of `None` from (1) to (3).

I'm also not sure why `defer=True` is a case. If we are going to make an API-breaking change (e.g. change the meaning of `timeout`), then we might as well choose the best API that we know of. When would you **not** want `ExecuteTrajectory` to return a future?

mkoval: Re: your latest comment.

What do you mean by "as it was executed"? Currently, we return the trajectory that was passed to the controller. This only differs from the input trajectory if the input was modified by the pre-processing steps (e.g. smoothed or re-timed). It does **not** reflect the actual trajectory that was followed by the manipulator.

For example, the a move-until-touch trajectory returned by `ExecuteTrajectory` will still be the full-length trajectory. Even if the actual trajectory executed on the robot immediately terminated due to an observed force.

psigen: I mean that we have switched to returning the trajectory which was actually followed by the manipulator over here (in our controller), and it is really really useful.

psigen: Didn't realize timeout was already overloaded.  We should special-case the 0 case then.

We don't want ExecuteTrajectory to return a future in situations where people just want to write a script that does a few things, in order.  Or to minimize modification to existing code.

Unless you want to seriously bust the API and use exclusively futures everywhere, but that seems like it will massively confuse people.

mkoval: We're already "seriously busting the API" if we're making `timeout = 0` immediately throw an exception. Also, there aren't very many cases where we actually use the return value from `ExecuteTrajectory`. The return value primarily exists for debugging purposes (i.e. in case the pre-processing code is broken).

What's the downside to using futures everywhere?

src/prpy/planning/base.py
psigen: Fixed in 921e212

src/prpy/planning/base.py
psigen: Fixed in 58d55c7.

src/prpy/planning/base.py
psigen: Fixed in 58d55c7.

src/prpy/planning/base.py
psigen: Fixed in 6b0636c.

src/prpy/planning/base.py
```diff
@@ -52,13 +52,30 @@ def __init__(self, func):
...
 
-        return traj
+            def call_planner():
+                try:
+                    planner_traj = self.func(instance, cloned_robot,
+                                             *args, **kw_args)
+                    traj = openravepy.RaveCreateTrajectory(
+                        env, planner_traj.GetXMLId()
+                    )
+                    traj.Clone(planner_traj, 0)
```
mkoval: You could use `prpy.rave.CloneTrajectory` here. Less code duplication. :smile:

src/prpy/planning/base.py
```diff
@@ -52,13 +52,30 @@ def __init__(self, func):
...
+                    return traj
+                finally:
+                    cloned_env.Unlock()
+
+            if defer is True:
```
mkoval: Should this be an explicit comparison against `True`? Or just a check for truthiness.

mkoval:
Tested. LGTM.

PR #44, opened 2015-02-09
Added fix for error caused by clone_env being set to None.

psigen:
This PR fixes the bug where Cloned(...) calls fail due to an internal reference to clone_env being None.

This PR also adds a Cloned(clone_env=...) parameter for use in threads to make cloning more thread-safe.


mkoval:
LGTM. @aaronjoh also confirms that the demo runs with this branch.

@pkv Are you going to add the method to `Environment` in a separate PR?

PR #46, opened 2015-02-10
Added numerous bugfixes for cloning and deferred planning.

psigen:
* Deferred planning now consistently returns trollius.futures.Future
* Fixed bug in robot PlanWrapper that caused deferred planning to terminate early.
* Cloned() references are now explicitly passed their clone environment.
* .Cloned() helper method added to environments created by Clone(env)
* Existing clone references consolidated to minimize Cloned() lookups.
* Added usage of CopyTrajectory helper instead of direct Clone calls where applicable.

### DIFF
--- a/src/prpy/base/manipulator.py
+++ b/src/prpy/base/manipulator.py
@@ -141,19 +141,25 @@ class Manipulator(openravepy.Robot.Manipulator):
         self.GetRobot().SetDOFAccelerationLimits(or_accel_limits)
 
     def _PlanWrapper(self, planning_method, args, kw_args):
-        from prpy.clone import Clone, Cloned
+        from prpy.clone import Clone
         robot = self.GetRobot()
-        with Clone(robot.GetEnv()):
-            Cloned(self).SetActive()
+        with Clone(robot.GetEnv()) as cloned_env:
+            cloned_env.Cloned(self).SetActive()
+            cloned_robot = cloned_env.Cloned(robot)
             cloned_args = copy.copy(kw_args)
             cloned_args['execute'] = False
-            cloned_traj = Cloned(robot)._PlanWrapper(planning_method, args, cloned_args)
+            cloned_traj = cloned_robot._PlanWrapper(planning_method,
+                                                    args, cloned_args)
 
             # Strip inactive DOFs from the trajectory.
-            config_spec = Cloned(robot).GetActiveConfigurationSpecification()
-            openravepy.planningutils.ConvertTrajectorySpecification(cloned_traj, config_spec)
-            traj = openravepy.RaveCreateTrajectory(robot.GetEnv(), cloned_traj.GetXMLId())
-            traj.Clone(cloned_traj, 0)
+            config_spec = cloned_robot.GetActiveConfigurationSpecification()
+            openravepy.planningutils.ConvertTrajectorySpecification(
+                cloned_traj, config_spec
+            )
+
+            # Copy the trajectory back to the original environment.
+            from ..util import CopyTrajectory
+            traj = CopyTrajectory(cloned_traj, env=robot.GetEnv())
 
         # Optionally execute the trajectory.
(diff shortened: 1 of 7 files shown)


src/prpy/base/manipulator.py
```diff
@@ -141,19 +141,25 @@ def SetVelocityLimits(self, velocity_limits, min_accel_time):
         self.GetRobot().SetDOFAccelerationLimits(or_accel_limits)
 
     def _PlanWrapper(self, planning_method, args, kw_args):
-        from prpy.clone import Clone, Cloned
+        from prpy.clone import Clone
         robot = self.GetRobot()
-        with Clone(robot.GetEnv()):
-            Cloned(self).SetActive()
+        with Clone(robot.GetEnv()) as cloned_env:
```
mkoval: I think we should pass `lock=True` here.

src/prpy/base/mico.py
```diff
@@ -139,21 +139,23 @@ def PlanToNamedConfiguration(self, name, execute=True, **kw_args):
         robot = self.GetRobot()
         saved_dof_indices, saved_dof_values = robot.configurations.get_configuration(name)
 
-        with Clone(robot.GetEnv()):
-            Cloned(self).SetActive()
-            arm_dof_indices = Cloned(robot).GetActiveDOFIndices()
-            arm_dof_values = Cloned(robot).GetActiveDOFValues()
+        with Clone(robot.GetEnv()) as cloned_env:
```
mkoval: There should also be a `lock=True` here.

src/prpy/base/manipulator.py
```diff
@@ -141,19 +141,25 @@ def SetVelocityLimits(self, velocity_limits, min_accel_time):
...
+            cloned_env.Cloned(self).SetActive()
```
mkoval: What was wrong with the previous `Cloned(self).SetActive()`. There are no threads here, so it should be unambiguous. Is this just a matter of preference?

src/prpy/base/mobilebase.py
```diff
@@ -145,27 +145,29 @@ def DriveStraightUntilForce(self, direction, velocity=0.1, force_threshold=3.0,
...
-        from prpy.clone import Clone, Cloned
-        
         robot = self.robot
-        with Clone(robot.GetEnv()):
-            Cloned(robot).SetActiveDOFs([], 
-                                        affine = openravepy.DOFAffine.X |
-                                        openravepy.DOFAffine.Y | openravepy.DOFAffine.RotationAxis)
-
-            cloned_traj = planning_method(Cloned(robot), *args, **kw_args);
+        with Clone(robot.GetEnv()) as cloned_env:
```
mkoval: There should also be a `lock=True` here.

src/prpy/base/robot.py
```diff
@@ -317,14 +320,26 @@ def postprocess_trajectory(traj, kw_args):
...
         # Return either the trajectory result or a future to the result.
         if 'defer' in kw_args and kw_args['defer'] is True:
-            from trollius.executor import get_default_executor
-            executor = kw_args.get('executor') or get_default_executor()
-            return executor.submit(defer_trajectory, result, kw_args)
+            import trollius
+
+            # Perform postprocessing on a future trajectory.
+            @trollius.coroutine
+            def defer_trajectory(traj_future, kw_args):
```
mkoval: I thought we mostly moved away from Trollius coroutines to Trollius futures. Is this an artifact of this branch being out of date? Or are we still using coroutines in some places?

mkoval:
This looks good to me, other than those missing locks.

I also wish you'd write documentation for #48. I'm still a bit confused myself by the way we ended up using Trollius. :smile:

src/prpy/base/manipulator.py
psigen: Nothing wrong here, just making it consistent.

PR #47, opened 2015-02-11
README documentation

mkoval:
I added a `README` to PrPy as part of our documentation effort.

@jeking04 Could you take a look at the changes? In particular, the new "Method Binding" section.


mkoval:
Also: Can @pkv add a short subsection to the "Planning" section describing `defer=True`?

I'm still a bit confused about the difference between `trollious.future` and `concurrent.future`, so I'd like to avoid writing this documentation myself.

PR #50, opened 2015-02-13
Workspace trajectory planner with Greedy IK

siddhss5:
Follows a workspace path exactly at robot.GetDOFResolutions.

### DIFF
--- a/src/prpy/planning/workspace.py
+++ b/src/prpy/planning/workspace.py
@@ -80,64 +80,62 @@ class GreedyIKPlanner(BasePlanner):
 
         return self.PlanWorkspacePath(robot, traj, timelimit)
 
-        @PlanningMethod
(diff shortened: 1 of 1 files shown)


siddhss5:
I just thought of a potential concern. 
Do we need to have some sort of `UpdateWorld` function call inside the loop right after we step the trajectory forward? For example, if we're opening a door, we need to first follow the arc of the circle but then also need to step the door forward so that the next IK check has the updated world to collision check against.
In general, do we have a programmatic way to do this? It seems like something that other planners might need too.

siddhss5:
@psigen and I had a chat about the above comment. He suggested we make this a separate discussion point and focus this PR on the planner. I agree.

mkoval:
The implementation looks good to me. The time stepping code is quite elegant! :smile:

@siddhss5 Could you make an issue for that discussion?

PR #51, opened 2015-02-14
Added bugfix for cloning grabbed objects.

psigen:
Due to a bug in OpenRAVE, cloned grabbed objects may have incorrect adjacency properties, causing them to not be evaluated correctly for self collisions (with the robot).  This bugfix forces cloned environments to regrab all objects, which resets these incorrect links.

### DIFF
--- a/src/prpy/clone.py
+++ b/src/prpy/clone.py
@@ -87,6 +87,24 @@ class Clone(object):
             # PrPy-annotated classes.
             setattr(self.clone_env, 'clone_parent', self.clone_parent)
 
+            # Due to a bug in the OpenRAVE clone API, we need to regrab
+            # objects in cloned environments because they might have
+            # incorrectly computed 'ignore' flags.
+            # TODO(pkv): Remove this block once OpenRAVE cloning is fixed.
+            for robot in self.clone_parent.GetRobots():
+                # Since the new ignore lists are computed from the current
+                # pose,  calling RegrabAll() from a pose that is in
+                # SelfCollision may incorrectly ignore collisions.
+                if robot.CheckSelfCollision():
+                    raise CloneException(
+                        'Unable to compute self-collisions correctly. '
+                        'Robot {:s} was cloned while in collision.'
+                        .format(robot.GetName())
+                    )
+                cloned_robot = self.clone_env.Cloned(robot)
+                with self.clone_env:
+                    cloned_robot.RegrabAll()
+
     def __enter__(self):
         if self.lock:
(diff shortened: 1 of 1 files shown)


mkoval:
This is likely related to https://github.com/rdiankov/openrave/issues/335. It's really unfortunate that we still have these mysterious cloning issues in OpenRAVE.

In any case, it's better to have the check than to silently fail. :+1:

sjavdani:
Ok, so I'm running into similar problems now (and related to #51). Looks like the following:

If you call RegrabAll, as #51  changed the code to do, the list of links to ignore for a grasped object are reset. More specifically, if I call robot.GetGrabbedInfio()[idx]._setRobotLinksToIgnore, the list is different then cloned_robot.GetGrabbedInfio()[idx]._setRobotLinksToIgnore. But if you do not call RegrabAll(), then this doesn't happen.

So regrab isn't setting the correct collision list. I don't see any way to manually modify this list. I'm not sure what regrab was helping with actually, since it is causing an error now, so I can't remove it and test to see if it doesn't break anything for anyone.

psigen:
This is extremely case-specific.  While calling `RegrabAll()` recreates the collision list _in your particular case_, *not* calling `RegrabAll()` can arbitrarily cause the list to change to be just plain invalid: i.e. it can have links which are neither adjacent nor non-adjacent, or _both_.  Why this happens is very mysterious.  See #255.

In any case, the robot links are set from scratch based on OpenRAVE's heuristic, which disables links that are in contact when the clone happens, as well as links flagged as adjacent in the robot XML/SRDF.  You can try moving the robot to a known-good pose where it will recompute the collisions that you want, then clone, then move it back in the cloned environment.

sjavdani:
I'm having trouble with a collision between the robot link and a grabbed
body, not two robot links. When the object was originally grabbed, I
specified a list of links that collision should be ignored with (all finger
links). This is the list that is not copied when RegrabAll() is called. In
this case, does your suggestion say I should make sure the fingers are
touching the object before RegrabAll() is called?

What I'm really confused about is why this **isnt** an issue more often,
since I'm guessing closing the hand and calling grab will often mean the
hand's collision mesh is colliding with the robots. I hadn't previously
needed to manually specify links to ignore before, so am I just doing
something else in a weird order that I don't usually do?

Also, I don't see an easy way to set this list of links to ignore through
openrave's python interface. Is there one? If so, seems like the best fix
is to manually make sure the cloned and original environments have these
same properties.

PR #52, opened 2015-02-16
Bugfix: IK Ranking and SimplifyTrajectory

psigen:
This PR adds handling of two special cases to planning utility functions `prpy.ik_ranking.NominalConfiguration` and `prpy.util.SimplifyTrajectory`.

* `SimplifyTrajectory` has been modified to gracefully return if passed a trajectory with only one waypoint.
* `NominalConfiguration` optionally takes a maximum allowable DOF range, which allows robots with fully redundant configurations (i.e. multiple rotation joints) to ignore IK configurations for which a closer solution must exist.

### DIFF
--- a/src/prpy/ik_ranking.py
+++ b/src/prpy/ik_ranking.py
@@ -53,13 +53,26 @@ def JointLimitAvoidance(robot, ik_solutions):
 
 
 class NominalConfiguration(object):
-    def __init__(self, q_nominal):
+    def __init__(self, q_nominal, max_deviation=2*numpy.pi):
         """
         Score IK solutions by their distance from a nominal configuration.
         @param q_nominal nominal configuration
+        @param max_deviation specify a maximum allowable per-joint deviation
+                             from the nominal configuration, default is 2*PI
         """
         self.q_nominal = q_nominal
+        self.max_deviation = max_deviation
 
     def __call__(self, robot, ik_solutions):
         assert ik_solutions.shape[1] == self.q_nominal.shape[0]
-        return numpy.linalg.norm(ik_solutions - self.q_nominal, axis=1)
+        L_2 = numpy.linalg.norm(ik_solutions - self.q_nominal,
+                                axis=1, ord=2)
+
+        # Ignore IK solutions that are more than the specified distance away.
+        # (This means a closer solution must exist!)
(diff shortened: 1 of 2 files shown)


src/prpy/ik_ranking.py
```diff
@@ -53,13 +53,26 @@ def JointLimitAvoidance(robot, ik_solutions):
 
 
 class NominalConfiguration(object):
-    def __init__(self, q_nominal):
+    def __init__(self, q_nominal, max_deviation=2*numpy.pi):
```
mkoval: I think this default will break on robots that have revolute joints with more than 360 degrees of rotation. This is the case on ADA, along with any other robots, e.g. Robonaut, that don't use slip rings. Is it ever possible that it is feasible to move to a configuration that is further away, but has the same end-effector pose?

If we're not convinced that this can never happen, I think the `max_deviation` parameter should default to infinity.

mkoval:
Looks good to me, besides the default value of `max_deviation`.

src/prpy/ik_ranking.py
psigen: This default actually helps quite a lot on multirotation or continuous rotation robots.

The only special case where you might need explicit multirotation is where the intermediate trajectory needs to wrap a full rotation to avoid an intermediate obstacle that partitions the configuration space (like a corkscrew maze), which is uncommon unless a pathological environment is constructed.

On the other hand, there are many case where multiple copies of the same good IK solution dominate the ranking (if there are multiple joints and (n-1) match your current configuration closely, 2*PI deviation in only one joint is easily ranked better than smaller deviations on multiple joints).

Since this duplication is relative to the power set of the underconstrained multirotational joints, this can quickly lead to having to scan hundreds of near identical solutions to find the top few actual candidate solutions.  This is actually specifically why I had to implement this check in the first place.

mkoval: This is exactly the case I was referring to:

> The only special case where you might need explicit multirotation is where the intermediate trajectory needs to wrap a full rotation to avoid an intermediate obstacle that partitions the configuration space 

However, I didn't think about the combinatorial effect that this could have when there are multiple multi-revolution joints. I buy your argument now. :smile:

PR #54, opened 2015-02-18
Added workspace planner to prpy.planning init.py

psigen:
This just adds the new workspace planner to `__init__.py` so it can be imported from `prpy.planning`.


mkoval:
The `GreedyIKPlanner` doesn't have any dependencies, so there's no harm in merging this. :+1:

PR #55, opened 2015-02-18
Fixed a number of bugs related to workspace planner.

psigen:
This PR addresses several major bugs unmasked by the workspace planner.
1. Fixed a bug in cloning an environment into itself
   (needed for recursive `@PlanningMethod`s)
2. Fixed a bug in incorrect formatting of RetimeTrajectory error messages.
3. Fixed numerous small issues in the workspace planner:
  * Returning a 1-waypoint trajectory when started in-contact with an object.
  * Fixed max_distance calculation error from missing `numpy.copy()`
  * Simplified some of the workspace planning logic.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -176,7 +176,7 @@ class Robot(openravepy.Robot):
                       PlannerStatus.InterruptedWithSolution]:
             return retimed_traj
         raise PrPyException("Path retimer failed with status '{:s}'"
-                            .format(status))
+                            .format(str(status)))
 
     def ExecuteTrajectory(self, traj, retime=True, timeout=None,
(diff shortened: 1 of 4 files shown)


mkoval:
I think we should print a warning if you try to clone an environment into itself. It seems very easy to run into aliasing issues like this. It also makes it easy to accidentally introduce a major bug in an otherwise innocuous typo:

```python
with Clone(env, into=env): # whoops, I meant into=env2
    # do stuff on env2, which is now in the incorrect state
```

We could either add an optional argument to disable the warning (e.g. `warn_on_same=False`) or simply require the caller to not call `Clone` in this case. Thoughts?

In any case, the rest of the pull request LGTM, so I am merging it. Thanks for the bug fixes. :smile:

PR #56, opened 2015-02-19
An optimization-based approach for computing joint velocities from twist

siddhss5:
I'm solving the problem of computing optimal joint velocities given a se(3) twist, and velocity and joint limits by formulating it as a quadratic optimization problem with box constraints. It turns out this is ridiculously fast to solve and implemented in SciPy as L-BFGS-B. Several magic incantations later, I have something that works remarkably fast and well.


mkoval:
Very cool implementation! We should switch to using this in `MKPlanner`.

PR #58, opened 2015-02-19
Revert "An optimization-based approach for computing joint velocities fr...

siddhss5:
...om twist"


mkoval:
I was able to merge #57 cleanly. I think this revert is unnecessary. Is it safe for me to close this PR and leave it un-merged?

PR #59, opened 2015-02-25
Added a few cleanups for syntax and simplicity.

psigen:
* Changed name capitalization to CamelCase.
* Removed active bodies list (pretty sure env does this automatically with the scene check)
* Rearranged q_curr for cleanliness.

### DIFF
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -38,12 +38,12 @@ import prpy.util
 logger = logging.getLogger('planning')
 
 
-class VectorfieldPlanner(BasePlanner):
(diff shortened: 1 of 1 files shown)


mkoval:
:+1: OpenRAVE handles the `IsDisabled` check inside the collision checker, so we shouldn't duplicate this logic ourselves. This logic also forces the collision checker to re-generate the broad phase tree multiple times, which could make collision checking significantly slower.

@siddhss5 Can you sign off on this before I merge it? I just want to verify that the active bodies list was not a workaround for a deeper issue.

siddhss5:
> I'm all good.
I copied that logic from MKPlanner. Serves me right for blindly copying.

PR #62, opened 2015-02-26
Fixed GreedyIKPlanner to use active DOFs

Stefanos19:


mkoval:
It looks like you missed one other reference to `GetDOFResolutions`. I believe [this line](https://github.com/personalrobotics/prpy/blob/Bugfix_GreedyIKPlanner/src/prpy/planning/workspace.py#L169) should also be updated to `GetActiveDOFResolutions`.

@Stefanos19 Can you make this change and verify that the planner still works?

PR #78, opened 2015-04-01
Tag trajectories with information necessary to control smoothing

mkoval:
This pull request adds the `GetTrajectoryTags` and `SetTrajectoryTags` functions to `prpy.util`. These functions attach an arbitrary dictionary of string key/value pairs to a trajectory by serializing it to XML and storing it the trajectory's description. This data is cloned and serialized/deserialized along with the trajectory object.

I use these functions to tag trajectories with metadata in planners. The motivation is to use these tags to control what type of shorcutting, smoothing, and/or timing should be applied to the trajectory before execution.

This is roughly the logic I am implementing in an upcoming pull request:

- `constrained`, e.g. `PlanToEndEffectorOffset` or `PlanToTSR` with a constraint TSR
    - the trajectory was planned to obey a constraint
    - do not smooth the piecewise linear path during post-processing (i.e. do not stop at waypoints)
- `optimized`, e.g. CHOMP or TrajOpt
   - the trajectory optimizes a cost function
   - treat the trajectory as a continuous function during retiming (i.e. do not stop at waypoints)
- nothing e.g. `PlanToConfiguration` from an RRT
   - shortcut the trajectory and smooth it during retiming (i.e. change the path)

The `@PlanningMethod` decorator also tags all trajectories with two attributes:

- `planner_name`: name of the planner that returned the trajectory (e.g. `OMPLPlanner`)
- `planning_method`: name of the method that returned the trajectory (e.g. `PlanToConfiguration`)

This can be used to implement any arbitrary, complicated rules that occasionally appear. For example, we may want to retain the timing on the output of `PlanToEndEffectorOffset` on `VectorFieldPlanner`, but not on `PlanToEndEffectorOffset` on CBiRRT.

@siddhss5 @jeking04 @cdellin @psigen Thoughts?

### DIFF
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -33,7 +33,7 @@ import functools
 import logging
 import openravepy
 from ..clone import Clone
-from ..util import CopyTrajectory, SetTrajectoryTags
+from ..util import CopyTrajectory, GetTrajectoryTags, SetTrajectoryTags
 
 logger = logging.getLogger('planning')
 
@@ -70,14 +70,16 @@ class PlanningMethod(object):
                 try:
                     planner_traj = self.func(instance, cloned_robot,
                                              *args, **kw_args)
-                    copied_traj =  CopyTrajectory(planner_traj, env=env)
 
                     # Tag the trajectory with the planner and planning method
-                    # used to generate it.
-                    SetTrajectoryTags(copied_traj, {
-                        'planner': instance.__class__.__name__,
-                        'planning_method': self.func.__name__,
-                    })
+                    # used to generate it. We don't overwrite these tags if
+                    # they already exist.
+                    tags = GetTrajectoryTags(planner_traj)
+                    tags.setdefault('planner', instance.__class__.__name__)
+                    tags.setdefault('planning_method', self.func.__name__)
+                    SetTrajectoryTags(planner_traj, tags, append=False)
+
+                    return CopyTrajectory(planner_traj, env=env)
 
                     return copied_traj
                 finally:
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -33,7 +33,7 @@ import functools
 import logging
 import openravepy
 from ..clone import Clone
-from ..util import CopyTrajectory
+from ..util import CopyTrajectory, SetTrajectoryTags
 
 logger = logging.getLogger('planning')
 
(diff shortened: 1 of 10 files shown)


siddhss5:
I **love** this change. Great work. Do we need a full refactor of `prpy/planning` once this is done?

PyryM:
For just key/value storage JSON has a number of advantages over XML: it is more compact, it parses faster, the parsing libraries are smaller, and it is more human-readable.

PyryM:
For example, (I don't know how to link to lines of code in a commit), with JSON this:
+    root_ele = lxml.etree.Element(_tag='tags')
+
+    for key, value in all_tags.iteritems():
+        lxml.etree.SubElement(
+            _parent=root_ele, _tag='tag',
+            attrib={'key': key, 'value': value}
+        )
+
+    description = lxml.etree.tostring(root_ele)
+ traj.SetDescription(description)

becomes
traj.SetDescription(json.dumps(all_tags))

and GetDescription becomes:
return json.loads(traj.GetDescription())

rather than the xml monstrosity it is now.

mkoval:
@PyryM I also prefer JSON. Unfortunately, OpenRAVE already serializes trajectories as XML. Serializing this as JSON would involve embedding a JSON literal inside a larger XML document.

XML is a pain to work with. It may still be worth it...

src/prpy/planning/base.py
```diff
@@ -70,7 +70,18 @@ def call_planner():
...
+                    # used to generate it. We don't overwrite these tags if
+                    # they already exist.
+                    tags = GetTrajectoryTags(planner_traj)
+                    tags.setdefault('planner', instance.__class__.__name__)
+                    tags.setdefault('planning_method', self.func.__name__)
+                    SetTrajectoryTags(planner_traj, tags, append=False)
+
                     return CopyTrajectory(planner_traj, env=env)
+
+                    return copied_traj
```
jeking04: I might be reading this wrong, but it looks like this returns twice and copied_traj is never set to anything.

mkoval: Fixed!

mkoval:
@PyryM I switched to JSON!

I'm holding off on the CHOMP cleanup in this pull request. There are a bunch of other PEP-8 fixes that we need to make in that module, so I'll do it as a separate commit.

mkoval:
@jeking04 I moved `CONSTRAINED` and `SMOOTH` into an enum.

PR #79, opened 2015-04-02
Cleaned up CloneBindings functions

mkoval:
@psigen caught a memory leak when cloning in PrPy. At least on HERB, this was caused by a circular reference created by the `ServoSimulator` in cloned environments. This made me realize that we're currently cloning a lot of things that are never used in cloned environments (e.g. controllers, simulators, some planners).

This pull request cleans up all of the `CloneBindings` functions in PrPy that are used on HERB:

- Reference the TSRLibrary from the parent environment.
- Reference the NamedConfigurations from the parent environment.
- Don't load ServoSimulatored in cloned environments.
- Don't load any controllers in cloned environments.
- Avoid calling `__init__` to prevent future nasty surprises.

### DIFF
--- a/src/prpy/base/endeffector.py
+++ b/src/prpy/base/endeffector.py
@@ -37,8 +37,9 @@ class EndEffector(openravepy.Robot.Link):
 
     def CloneBindings(self, parent):
         from ..clone import Cloned
-        EndEffector.__init__(self, Cloned(parent.manipulator,
-                                          into=self.GetParent().GetEnv()))
+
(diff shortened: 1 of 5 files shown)


mkoval:
@jeking04 Let me know if there are any issues with this. If not, I'll merge it. This should explain the enormous number of threads we saw while running the demo a few weeks ago. Each clone creates four additional threads that are never cleaned up :smile:

psigen:
Yeah, this fixes the thread issue that we had as well.  Nice work! :+1:

PR #80, opened 2015-04-02
feature added to avoid joint limit with ComputeJointVelocityFromTwist

sjavdani:
Added a new objective function, to be used for ComputeJointVelocityFromTwist, which adds a cost and gradient to avoid joint limits. Also modified ComputeJointVelocityFromTwist in order to make it have a hard constraint near a limit (was previously hard constrained at 0, now constrained at whatever joint_limit_tolerance is)

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -428,35 +428,29 @@ class Timer(object):
         return self.end - self.start
 
 
-def quadraticPlusJointLimitObjective(dq, *args):
+def quadraticPlusJointLimitObjective(dq, J, dx, q, q_min, q_max, delta_joint_penalty=5e-1, lambda_dqdist=0.01, *args):
     '''
-    Quadratic objective function for SciPy's optimization.
+    Quadratic plus joint limit avoidance objective function for SciPy's optimization.
     @param dq joint velocity
-    @param args[0] Jacobian
-    @param args[1] desired twist
-    @param args[2] current joint values
-    @param args[3] lower joint limit
-    @param args[4] upper joint limit
+    @param J Jacobian
+    @param dx desired twist
+    @param q current joint values
+    @param q_min lower joint limit
+    @param q_max upper joint limit
+    @param delta_joint_penalty distance from limit with penality
+    @param lamdbda_dqdist weighting for joint limit penalty
     '''
     
-    objective, gradient = quadraticObjective(dq, *args)
-    q = args[2]
-    q_min = args[3]
-    q_max = args[4]
+    #compute quadratic distance part
+    objective, gradient = quadraticObjective(dq, J, dx)
 
-    #when joint is less than this distance away from limit, start penalizing
-    delta_j = 3e-1
-
-    qdiff_lower = delta_j - (q-q_min)
-    qdiff_upper = delta_j - (q_max-q)
-
+    #add penalty for joint limit avoidance
+    qdiff_lower = delta_joint_penalty - (q-q_min)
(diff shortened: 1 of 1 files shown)


src/prpy/util.py
```diff
@@ -428,11 +428,47 @@ def get_duration(self):
...
+    @param args[4] upper joint limit
+    '''
+    
+    objective, gradient = quadraticObjective(dq, *args)
+    q = args[2]
+    q_min = args[3]
+    q_max = args[4]
+
+    #when joint is less than this distance away from limit, start penalizing
+    delta_j = 3e-1
```
siddhss5: Why is this hardcoded?

src/prpy/util.py
```diff
@@ -428,11 +428,47 @@ def get_duration(self):
...
+    qdiff_lower = delta_j - (q-q_min)
+    qdiff_upper = delta_j - (q_max-q)
+
+
+    dq_target = [diff_lower if diff_lower > 0. else
+                 -diff_upper if diff_upper > 0. else
+                 0. for diff_lower, diff_upper in zip(qdiff_lower, qdiff_upper)]
+
+    #constant for weighing cost of quadratic vs. being diff from bounds
+    lambda_dqdist = 0.01
```
siddhss5: Why is this hardcoded?

src/prpy/util.py
```diff
@@ -446,7 +482,8 @@ def quadraticObjective(dq, *args):
 
 
 def ComputeJointVelocityFromTwist(robot, twist,
-                                  objective=quadraticObjective,
+                                  objective=quadraticPlusJointLimitObjective,
```
siddhss5: I think `quadraticObjective` should be default in `prpy.util`. There are way too many parameters in `quadraticPlusJointLimitObjective` for this to be easily generalizable to all robots. If you feel like it, you should use the other objective with specially tuned parameters within a specific `robot.py`.

src/prpy/util.py
```diff
@@ -478,18 +515,20 @@ def ComputeJointVelocityFromTwist(robot, twist,
...
     if dq_init is None:
         dq_init = robot.GetActiveDOFVelocities()
 
     opt = scipy.optimize.fmin_l_bfgs_b(objective, dq_init, fprime=None,
-                                       args=(jacobian_active, twist_active),
+                                       args=(jacobian_active, twist_active, q_curr, q_min, q_max),
                                        bounds=dq_bounds, approx_grad=False)
 
     dq_opt = opt[0]
+#    if opt[1] > 0:
```
siddhss5: This has been removed in a downstream PR.

siddhss5:
I'm concerned about this change. We use `VectorField` extensively and it's fast because `ComputeJointVelocityFromTwist` is fast. I'd be wary about adding more optimization logic. I understand stepwise performance can be better but I'm not sure it has been tested enough to warrant a wholesale switch.

mkoval:
I do agree that we should leave `quadraticObjective` as the default and move this to `prpy.util`. It does take a lot of parameters that may not have sane default values.

sjavdani:
Ok, changed the default to be quadratic objective. In order to make it so you can change the constants, I also changed it so some of the arguments had to be specified for both quadratic and quadratic+joint objectives.

Note that, compared to before, it has a different tolerance at which joint velocities are forced to be zero. I found this limit to work much better in practice - Sidd, if this messed up other stuff, feel free to make the constant (joint_limit_tolerance) smaller.

mkoval:
Those changes look good to me. :+1: 

@siddhss5 Any other comments? Or can I merge?

PR #81, opened 2015-04-02
Added new MethodMask and FirstSupported meta-planners

mkoval:
- Added the `FirstSupported` meta-planner. This meta-planner operates on a list of planners and calls the first planner in the list that supports the desired planning method.
- Added the `MethodMask` meta-planner. This meta-planner operates on a single planner by only allowing access to a subset of its planning methods.
- Added support for explicitly passing a delegate planner to:
    - IKPlanner
    - NamedPlanner
    - TSRPlanner
- Modified `TSRPlanne`r to raise an UnsupportedPlanningError when it receives unsupported TSRs. This is necessary to trigger the fallback behavior implemented in the Fallback meta-planner.

### DIFF
--- a/src/prpy/planning/__init__.py
+++ b/src/prpy/planning/__init__.py
@@ -29,8 +29,8 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 from base import (
-    Fallback,
-    Only,
+    FirstSupported,
+    MethodMask,
     Planner,
     PlanningError,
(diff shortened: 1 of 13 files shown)


mkoval:
Signed off by @jeking04.

@psigen I implemented `PlanToIK` and `TSRPlanner`, but left it temporarily disabled. Why do you think this is a bad idea?

PR #82, opened 2015-04-02
Bugfix/servo simulation

sjavdani:
A tiny bit of bookkeeping so that the servo simulation sleeps for a variable amount of time, depending on how long was spent on computation


mkoval:
This change looks good. I am just waiting on merging this until we merge #80 because this pull request contains some of those commits as well.

PR #84, opened 2015-04-02
Fixing a few calls in VectorField planner that need to reference active ...

### LINKED ISSUE #83: VectorField planner fails on herb
On line 152 we need to call GetActiveDOFVelocityLimits rather than GetDOFVelocityLimits. The same is likely true in a few places in FollowVetorField and in PlanToEndEffectorPose.

jeking04:
...dofs.

This fixes #83


siddhss5:
Sorry about that! I think there's a call to GetDOFResolutions instead of GetActiveDOFResolutions that needs to be changed too.

Issue #73, opened 2015-03-25
OMPLSimplifier uses timeout instead of timelimit as the planning budget


mkoval:
All of our planners use the `timelimit` keyword argument for the planning time budget. The `timeout` keyword is reserved as a timeout on trajectory execution. `OMPLSimplifier` breaks this convention and causes problems when passing `timeout=0` along with `execute=True`. See https://github.com/personalrobotics/herbpy/issues/12.

This raises a deeper issue: if we use `timelimit` on `OMPLSimplifier`, it will no longer be possible to pass separate time limits for planning and path simplification. We typically use a much longer time limit for planning (e.g. 30 seconds) than for simplification (e.g. 5 seconds), so this is undesirable.

The quick fix is to use a different keyword for this time limit, e.g. `smoothing_timelimit`. Alternatively, maybe we should find a more principled method of setting this argument when we refactor the way we call post-processing functions.

@psigen Thoughts?

cdellin:
I know I've complained about this before, but I just want to interject here for an outsider's perspective. It doesn't make sense to me that a method called "PlanToXxx()" would ever execute anything (I would expect it to return a trajectory object).  Furthermore, I would expect a parameter named "timelimit" or "timeout" on a "PlanToXxx()" method to affect the planning (not execution).

I would suggest a different set of calls, perhaps "MoveToXxx()", which internally do planning and execution for you.  For these methods, I would expect parameters to be something like "planning_timeout" and/or "execution_timeout".  In the case you raise above, perhaps an explicit "smoothing_timeout" would be useful as you mentioned.

If this is not possible, then *at least* please rename the existing "timeout" parameter to be "execution_timeout" so that it's not extremely confusing.  `</rant>`

mkoval:
@cdellin I think that is an orthogonal issue. As you mentioned, would have the same name conflict in `MoveToXxx`. We will still have to settle on a new naming convention.

That being said, I'm open to that suggestion. I also don't like how `PlanToConfiguration` behaves completely differently depending upon whether `execute=True` is passed. We can't quite use your suggestion (replacing `Plan` with `Move`) because some planning methods, e.g. `OptimizeTrajectory` do not start with `Plan`. However, we could do something similar by appending a suffix to the method name; e.g. `PlanToXxxAndExecute`.

cdellin:
@mkoval The main thrust of my complaint is that when I read a method call like:

    robot.PlanToHome(timeout=10.0, numretries=3, inparallel=True)

[...]

PR #95, opened 2015-04-05
Trajectory timing/smoothing refactor 2.0

mkoval:
This pull request finishes (for now...) the trajectory timing/smooth refactor that I started in #64 and #78. It also addresses the issue that @garthz raised in #73.

Here's a quick summary:

- Added a new `PostProcessPath` function with three different post-processing pipelines:
    1. for smooth trajectories (e.g. CHOMP), approximate the trajectory as a continuous curve
    2. for constrained trajectories, stop at every waypoint and exactly follow the geometric path
    3. otherwise, by default, shortcut the path and simultaneously smooth/retime it
- `PostProcessPath` takes separate `shortcut_`, `smoothing_`, and `retiming_options `dicts (as per #73)
- We determine which pipeline to use (i, ii, or iii) using the tags on the trajectory (#78)
- **Note:** We current have no method of doing (i), so it temporarily falls through to (ii) and (iii)
- Modified `ExecutePath` to call `PostProcessPath`, followed by `ExecuteTrajectory`
- Added docstrings for `PostProcessPath`, `ExecutePath`, and `ExecuteTrajectory`

`PostProcessPath` (and, thus, `ExecutePath`) do not correctly handle base trajectories. I want to get @jeking04's input on this before I proceed, but I expect to:

1. Check if any affine DOFs are in the path passed to `PostProcessPath`; if:
    - *Only regular DOFs* use the current behavior
    - *Only affine DOFs* call `RetimeAffineDOFTrajectory`
    - *Both are present* raise an exception
3. Call `ExecuteTrajectory` as normal (I believe this already handles base trajectories)

Finally, there a few obvious next steps. I will create separate issues for these:

- Implement a method retiming smooth paths; i.e. case (i)
- Tag the trajectory with constraint TSRs
- If a trajectory is tagged with constraint TSRs, use CBiRRT to do the smoothing/timing

@siddhss5 @psigen @jeking04 Thoughts?

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -269,6 +269,8 @@ class Robot(openravepy.Robot):
                     'Setting robot "%s" DOFs %s as active for post-processing.',
                     cloned_robot.GetName(), list(dof_indices))
 
+                # TODO: Handle a affine DOF trajectories for the base.
+
                 # Directly compute a timing of smooth trajectories.
                 if smooth:
                     logger.warning(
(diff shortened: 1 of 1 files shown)


mkoval:
We were just waiting for @jeking04 to collect data before merging this. I am going ahead and doing the merge now.

Issue #93, opened 2015-04-05
Mico, MicoRobot, and MicoHand contain meaningless OWD-specific code


mkoval:
It looks like `Mico`, `MicoRobot`, and `MicoHand` were directly copied-and-pasted from HERB with, in many cases, no modification. There are a lot of HERB and OWD-specific vestiges in these files that don't make any sense on ADA.

For example...
- `MicoHand` has:
    - `spread
    - `ResetHand()`
    - `GetState()`
- `MicoRobot` has:
    - `ExecuteTrajectory()` has logic to deal with multiple manipulators and a mobile base
    - a bunch of HERB-specific code commented out
- `Mico` has:
    - `Servo()`, as implemented for OWD (is this the same on the MICO?)
    - `GetTrajectoryStatus()`
    - `ClearTrajectoryStatus()`
    - `SetVelocityLimits()`

mkoval:
Fixed in #97.

Issue #91, opened 2015-04-05
CloneBindings on Ada will cause a memory leak


mkoval:
The current [implementation of `CloneBindings`](https://github.com/personalrobotics/prpy/blob/master/src/prpy/base/mico.py#L128) on Ada directly calls `__init__(sim=True)`. As we recently found out on HERB (https://github.com/personalrobotics/prpy/pull/79), this is a bad idea because it can lead to inadvertently instantiating more than you originally intended. In some cases, this can cause a memory leak.

In fact, this **is causing a memory leak now** because a `ServoSimulator` is being created in the cloned environment. The `ServoSimulator` does not, currently, properly clean up after itself in cloned environments (see https://github.com/personalrobotics/prpy/issues/87).

Stefanos19:
What does the CloneBindings do? Should I put a pass as in https://github.com/personalrobotics/prpy/blob/master/src/prpy/base/manipulator.py#L39 ?

mkoval:
`CloneBindings` is supposed to copy any PrPy-related metadata associated with the robot. This includes `robot.planner`, `robot.tsrlibrary`, `robot.right_arm`, etc. You can read more [in the PrPy documentation](https://github.com/personalrobotics/prpy#cloning-bound-subclasses).

As a rough guideline, `CloneBindings` should:

1. call the `CloneBindings` function of the superclass (in this case, `prpy.base.Robot`)
2. set any fields that are set by default in `__init__`

In some cases you'll need to create new objects (e.g. `robot.right_arm`, an IK solver). In other cases, you can reference the same objects associated with the parent class (e.g. `robot.planner` or `robot.tsrlibrary`). This depends entirely on whether the field is coupled to a single environment or not.

Stefanos19:
Fixed in https://github.com/personalrobotics/prpy/pull/97

PR #97, opened 2015-04-06
Mico Refactor

### LINKED ISSUE #90: Ada should use NamedPlanner; not re-implement PlanToNamedConfiguration
The `Mico` class in PrPy currently [implements `PlanToNamedConfiguration`](https://github.com/personalrobotics/prpy/blob/master/src/prpy/base/mico.py#L131). We have the [`NamedPlanner` helper class](https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/named.py#L33) to avoid having to re-implement this on each robot. We should switch to this on ADA.

### LINKED ISSUE #92: Ada has no IK solver loaded by default
There is no IK solver loaded by default on Ada. To replicate, run `robot.arm.GetIkSolver()` after calling `adapy.initialize()`. We should setup `or_nlopt_ik` with sane parameters by default in the `Mico` class. For planning to work in cloned environments, this needs to be done in both `__init__` and `CloneBindings`.

mkoval:
I started a major refactor of the Mico code in PrPy to address the issues raised in #90, #91, #92, #93, and others.

Currently, I have:

- Deleted `MicoRobot` because it the file was empty once I fixed #90 and #93
- Removed a bunch of OWD-specific code from `Mico` and `Micohand` (#93)
- Load `or_nlopt_ik` as the default IK solver in `Mico` (#92)
- Fixed all of the `CloneBindings` calls to not leak memory (#91)

This will most likely **not work** on the robot (or maybe even in simulation) in the current form. I still need to sort out a better way of interfacing with the controller. I just created this pull request to have a place to put some comments.


mkoval:
We successfully ran the demo using this branch, so I am merging it. Most of the changes are in files specific to ADA, so I don't expect this to break anything on HERB.

@jeking04 Let me know if you run into any problems with `master` after I merge this.

Issue #99, opened 2015-04-06
VectorField planner fails for Ada


Stefanos19:
The planner throws an error: ```Deviated from straight line constraint```
Removing the error check results in a curved trajectory

siddhss5:
Not sure if this is a bug really. I instrumented the planner and it seems to be trying to follow the vector field as much as it can. I wonder if this is incorrect:
```
self.robot.GetDOFVelocityLimits()
Out[2]: 
array([ 0.87266462,  0.87266462,  0.87266462,  2.  ,  2.  , 2.  ,  0.78539816,  0.78539816])
```
The [spec sheet] (http://kinovarobotics.com/wp-content/uploads/2015/01/Mico-Specification-Sheet-v1.0.8.pdf) seems to suggest that the `large actuators` have a limit of `9 RPM` which is `0.942 rad/s`. and the `small actuators` have a limit of `18 RPM`.

Stefanos19:
I see. The reason I am a bit suspicious is that I have not been able to have the planner succeed for Ada, even for small displacements (< 0.03).

psigen:
@Stefanos19: Can you make a small test script that just loads Ada into an empty environment, and then fails to plan a 1cm pure straight-line path using just the VectorFieldPlanner? 

@mkoval can show you how to directly call the planner if you need help.

mkoval:
I've actually never seen `VectorFieldPlanner` succeed on HERB, either. I did this call in the `home` configuration:

    robot.vectorfield_planner.PlanToEndEffectorOffset(robot, numpy.array([0,0,1.]), 0.01)

It throws this exception:

> PlanningError: Step size too small, unable to progress

I think @jeking04 ran into this issue as well and had to change something about the the `dt_multiplier` parameter to get it to work.

Stefanos19:
@psgen: I updated a script that tests  the vectorfield planner https://github.com/personalrobotics/ada_cbirrt_test/blob/master/src/vectorfield_planner_test.py
I get either the `Deviated from straight line constraint` or the ` Step size too small, unable to progress` error.

siddhss5:
Looks like some of my _enhancements_ to the planner broke it. Sorry. I think some of these issues should be fixed now. Let me know. I'm at DARPA away from a real computer.

Stefanos19:
With the latest changes, it works for displacements up to 5cm, but then it will throw the `deciated from straight line constraint` error.

mkoval:
[...]

PR #108, opened 2015-04-09
Fixed two bugs in vectorfield planner

siddhss5:
This PR references https://github.com/personalrobotics/prpy/issues/99
Fixed two bugs: 
1. Missing `abs`
2. Changed default `dt_multiplier` to 1.01 so that `numsteps` floors to 1 by default.


mkoval:
Yes, it works now! And it's super fast!

Issue #114, opened 2015-04-11
Iksolver reset to none


Stefanos19:
This is likely the same issue as https://github.com/personalrobotics/prpy/issues/111, but it appears also in the master version of prpy. After running [Jen's branch] (https://github.com/personalrobotics/ada_meal_scenario/tree/feature/action_framework) once, the iksolver is reset to none probably in some clone-bindings, and the planners that use an iksolver fail.

mkoval:
Mostly a note to self...the problem is that CloneBindings only gets called
the first time an environment is used as a clone target. We need to
manually clear parent pointers in the target environment in prpy.Clone.

Stefanos19:
Resolved with https://github.com/personalrobotics/prpy/commits/bugfix/CloneUserData

PR #115, opened 2015-04-11
Clear UserData in prpy.Clone

### LINKED ISSUE #111: feature/MICORefactor: IkSolver is reset to None
After a few executions, IkSolver seems to be reset to none, possibly in some of the clone-bindings. This happens even when I explicitly set the IkSolver to NloptIK in the bite_serving code, before calling a planner.

mkoval:
OpenRAVE's `Clone` function does not clear `UserData` on the target environment. We use `UserData` inside `InstanceDeduplicator` to associate objects between environments. This was causing us to only call `CloneBindings` the first time we clone into a target environment.

This should fix #111 and #114.

@Stefanos19 @jeking04: Can you try this and let me know?

### DIFF
--- a/src/prpy/clone.py
+++ b/src/prpy/clone.py
@@ -83,6 +83,11 @@ class Clone(object):
 
         # Actually clone.
         with self.clone_env:
+            # Clear user-data. Otherwise, cloning into into the same target
+            # environment multiple times may not cause CloneBindings to get
+            # called again.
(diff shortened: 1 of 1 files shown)


Stefanos19:
@mkoval: This appears to have fixed it, thanks a bunch! :) (Although, that's actually hard to tell, since this bug did not show up today after restarting the pc and running the prpy master version)

PR #120, opened 2015-04-21
Bugfix: IK GIL release

psigen:
This PR defaults FindIK calls in the planners to release the GIL while they execute.  This prevents thread locking during IK sampling.


mkoval:
Tested on HERB. :+1:

PR #121, opened 2015-04-24
added kwargs to ExecuteTrajectory and PostProcessPath

rachelholladay:
In the pipeline ExecuteTrajectory and PostProcessPath where both missing **kwargs

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -184,7 +184,7 @@ class Robot(openravepy.Robot):
     def PostProcessPath(self, path, defer=False, executor=None,
                         constrained=None, smooth=None, default_timelimit=0.5,
                         shortcut_options=None, smoothing_options=None,
-                        retiming_options=None):
+                        retiming_options=None, **kwargs):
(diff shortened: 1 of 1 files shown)


mkoval:
This is definitely a bug. The question is only whether we should fix it by: (1) adding `**kwargs` to `ExecuteTrajectory` and `PostProcessPath` or (2) not forwarding `**kwargs` to these functions in `_PlanWrapper`. I don't think we ever rely on this behavior, except perhaps for passing `timeout=0`, so I am in favor of (2).

@jeking04 Would this have any negative implications for the table clearing demo?

psigen:
Isn't it problematic that removing kwargs would leave *no* way of
specifying custom post-processing options?  What if I want to run the
planning stack but change one of the postprocessing timeouts (like a
smoothing_option) because it takes too long.  What are my options there?

mkoval:
You always have the option of calling `PostProcessPath` manually. Although, I do see your point. There isn't any harm in forwarding these arguments now that we use explicit `dict`'s to pass arguments to the sub-components of `PostProcessPath`. I am merging this pull request.

PR #123, opened 2015-04-24
Bugfix: PostprocessPath params

psigen:
This PR fixes two argument issues in the PostprocessPath() stack, one dealing with handling `GetTrajectoryTags()` errors gracefully and the other dealing with catching invalid `defer` arguments earlier.

See commit messages for more details.


mkoval:
This is definitely an improvement. I am concerned that this change will allow `defer` to silently eat a positional argument and still produce the desired output.

I made two changes to your pull request:

1. Print a warning if `defer` is not `True` or `False`.
2. Print a warning in `GetTrajectoryTags` if `GetDescription()` is non-empty, but contains invalid JSON.

I am merging this pull request with these changes. Feel free to re-open it if you disagree with them.

psigen:
Question:  If a string is null, isn't that still a trajectory tags error? An empty string isn't valid JSON either, hence it also throws an error.   I agree with printing a warning, but why special case the empty string?

PR #124, opened 2015-04-25
Restructured defer fixes to raise exception.

psigen:
Instead of printing a warning, this restructures the `defer` argument checking to raise an exception if an invalid value has been provided.

This seems more reasonable than just printing a warning.


mkoval:
:+1:

Issue #45, opened 2015-02-10
Increase granularity of exceptions thrown in PlanTo functions


ablasdel:
It would be a nice enhancement if, when the planners fail, if we can get more information about why it failed. 

For example:
Start configuration in collision
End configuration in collision
No plan found between Start/End

This would make it much easier to know how to react to failures in planning.

mkoval:
To track our progress on this:

- @mkoval added collision exceptions to `VectorFieldPlanner` in #184
- @psigen added exceptions for precondition violations in #196 
- @lgw903 added specific start and goal collision exceptions to `SnapPlanner` and `GreedyIKPlanner` in #205

mkoval:
This has been largely addressed by #184, #196, and #205.

PR #132, opened 2015-05-01
Added more fine-grained planning exceptions

mkoval:
This is a first step towards #45.

I added a few subclasses to `PlanningError`. This includes a `CollisionPlanningError` that makes it easy to format a readable exception from a `CollisionReport` structure; e.g.

```
CollisionPlanningError: Terminated early: Detected collision: <table, table> x <herb, /right/finger2_2>.
```

For now, I only raise these exceptions in `GreedyIkPlanner` and `VectorFieldPlanner`. I likely missed many other places where we can throw a more specific exception.

@psigen @jeking04 @siddhss5 Thoughts?


mkoval:
No one protested, so I am merging this.

PR #139, opened 2015-06-04
Changed loggers to use '__name__' instead of explicit paths.

psigen:
Since the predominant log convention was to use module name anyway,
it is simpler to simply dynamically query this instead of manually
specifying it each time.

This also fixes a number of typos and inconsistencies in the log
naming, due to the above.


mkoval:
:+1: We should have been doing this for a long time.

PR #140, opened 2015-06-04
Bugfix: cloning race condition

psigen:
This is a hotfix/patch for a bizarre race condition issue in which the cloned robot does not match the original robot while inside the cloning operation.  I really don't want to merge this, but I have no idea what could be causing this problem and it does happen semi-frequently.

Any ideas @mkoval @jeking04 ?

### DIFF
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -79,14 +79,18 @@ class PlanningMethod(object):
             joint_values[1] = cloned_robot.GetActiveDOFValues()
 
             # Check for mismatches in the cloning and hackily reset them.
+            # (This is due to a possible bug in OpenRAVE environment cloning where in
+            # certain situations, the Active DOF ordering and values do not match the
+            # parent environment.  It seems to be exacerbated by multirotation joints,
+            # but the exact cause and repeatability is unclear at this point.)
             if not numpy.array_equal(joint_indices[0], joint_indices[1]):
-                logger.info("Cloned Active DOF index mismatch: %s != %s",
+                logger.warn("Cloned Active DOF index mismatch: %s != %s",
                             str(joint_indices[0]),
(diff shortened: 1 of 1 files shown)


mkoval:
@psigen Do you still have this workaround in place on your system? We haven't observed this on HERB yet (that I know of), so I am very perplexed by this.

psigen:
Not only do we still have the workaround in place on our system, we see it kick in (via the logged warnings) about 1/20 clones.

mkoval:
In that case, I am happy to merge this. I haven't noticed the behavior on HERB, but perhaps we are just not aware of it. I only have two requests:

1. Change the `info` log message to `warning`s.
2. Add a comment above your addition that explains why this is necessary and links to this pull request.

psigen:
I am OK with doing the above, but I still really don't understand exactly why this happens, or if it could happen elsewhere...  Any insights on stuff I could check out and report on?

mkoval:
I have no idea. Are you sure you're diligently locking the environment *everywhere* you call non-threadsafe methods? Does it happen even if you don't have a viewer attached?

Two particularly hairy parts of OpenRAVE are mimic joints and joints that support more than 360 degrees of revolution. Both involve a lot of complex internal state. Are the issues isolated to either of those situations?

psigen:
Hmm, all good ideas, I'll do some tests around those particular areas this week.

mkoval:
@psigen Can you make those two requested changes first? We're about to do a `prpy` release and I'd like to get these changes in beforehand.

> 1. Change the info log message to warnings.
> 2. Add a comment above your addition that explains why this is necessary and links to this pull request.

psigen:
Ok, I think I addressed those comments..

mkoval:
:+1: Thanks!

PR #141, opened 2015-06-04
Changed default 'execute' behavior to False.

psigen:
Currently, it is very scary that calling a Plan...() function anywhere in the code without 'execute=False' can cause a robot to start moving and doing things.

It does not seem at all reasonable that this is the default behavior, especially now that we are increasingly using pipelined planning calls and post processing trajectories heavily before execution in demos-- where execute=True is exactly the wrong thing to do.

We should not have to pass a non-default argument to every single planning call to keep the robot from moving.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -570,7 +570,7 @@ class Robot(openravepy.Robot):
                 postprocess_trajectory(traj)
 
                 # Optionally execute the trajectory.
-                if kw_args.get('execute', True):
+                if kw_args.get('execute', False):
                     # We know defer = True if we're in this function, so we
                     # don't have to set it explicitly.
                     traj = yield trollius.From(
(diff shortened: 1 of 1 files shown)


psigen:
Hoping @cdellin and @jeking04 can back me up on this.

jeking04:
I agree that it is probably the right thing to do to have ```execute=False``` be the default behavior, as I would not expect a call to Plan...() to do anything more than plan.

In practice the change might be a bit of a pain because we have to upgrade all current calls where we expect execution to pass this flag.  On the positive side, in much of our current demo we are already passing ```execute=False``` so its probably not that much code that will need changed. Maybe an action in the action library, and some older demos that are probably broken in much bigger ways anyway.

In summary, I agree with this change.

mkoval:
I'm fine with this change. I'm also a bit concerned about the amount of code we'll have to update if we make this change.

psigen:
Can I merge this then?  Does someone need to update some code somewhere?

mkoval:
Here comes the breakage...

PR #144, opened 2015-06-04
Added support for environment (de)serialization

mkoval:
This pull request adds `serialize` and `deserialize` functions in the `prpy.serialization` module. These functions operate an an OpenRAVE environment by serializing to and deserializing from a `dict`. This `dict` only contains primitive types and be written to disk, e.g. using `pickle`, `json`, or `yaml`.


mkoval:
Signed off by @jeking04. :smile:

PR #145, opened 2015-06-05
Added forwarding of kwargs to TSRPlanner's delegate planner.

psigen:
This fixes an issue where it is impossible to pass general planner arguments through the TSRPlanner.  This is necessary for cases when you need to specify any global metaparameters to the delegate planner.

It avoids weird behavior where arguments that seem like they should have some effect are ignored (because the TSR metaplanner ate them).


psigen:
Can this be merged?

mkoval:
LGTM.

PR #146, opened 2015-06-11
Added proper checking for trajectory timing and length.

psigen:
This adds several checks to the ExecuteTrajectory function:
1) there must be at least one waypoint.
2) the first waypoint must match the robot's current config
   (within joint DOF resolutions)
3) if there is only one waypoint, succeed after the first two checks.
4) if there are multiple waypoints, the duration must be nonzero.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -421,7 +421,7 @@ class Robot(openravepy.Robot):
         will be raised). If timeout is a float (including timeout = 0), this
         function will return None once the timeout has ellapsed, even if the
         trajectory is still being executed.
-        
+
         NOTE: We suggest that you either use timeout=None or defer=True. If
         trajectory execution times out, there is no way to tell whether
         execution was successful or not. Other values of timeout are only
@@ -449,7 +449,13 @@ class Robot(openravepy.Robot):
 
         # If there was only one waypoint, at this point we are done!
         if traj.GetNumWaypoints() == 1:
-            return traj
+            if defer is True:
+                import trollius
+                future = trollius.Future()
+                future.set_result(traj)
+                return future
+            else:
+                return traj
 
         # Verify that the trajectory is timed.
         if traj.GetDuration() <= 0.0:
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -29,8 +29,7 @@
 # POSSIBILITY OF SUCH DAMAGE.
 
 import functools, logging, openravepy, numpy
-import prpy.util
-from .. import bind, named_config, planning, util
+from .. import bind, named_config, exceptions, util
 from ..clone import Clone, Cloned
 from ..tsr.tsrlibrary import TSRLibrary
 from ..planning.base import Sequence, Tags
@@ -281,7 +280,7 @@ class Robot(openravepy.Robot):
                 # Extract active DOFs from teh trajectory and set them as active.
                 dof_indices, _ = cspec.ExtractUsedIndices(self)
 
-                if prpy.util.HasAffineDOFs(cspec):
+                if util.HasAffineDOFs(cspec):
                     affine_dofs = (DOFAffine.X | DOFAffine.Y | DOFAffine.RotationAxis)
                     
                     # Bug in OpenRAVE ExtractUsedIndices function makes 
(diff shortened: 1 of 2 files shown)


src/prpy/base/robot.py
```diff
@@ -438,18 +437,33 @@ def ExecuteTrajectory(self, traj, defer=False, timeout=None, period=0.01, **kwar
...
+
+        # Check that the current configuration of the robot matches the
+        # initial configuration specified by the trajectory.
+        if not util.IsAtTrajectoryStart(self, traj):
+            raise exceptions.TrajectoryAborted(
+                'Trajectory started from different configuration than robot.')
+
+        # If there was only one waypoint, at this point we are done!
+        if traj.GetNumWaypoints() == 1:
+            return traj
```
mkoval: This will break if `defer=True`. In that case, you should return a `Future` that yields `traj`, instead of returning `traj` itself.

src/prpy/util.py
```diff
@@ -693,3 +693,44 @@ def FindCatkinResource(package, relative_path):
...
+
+    @param robot: the robot whose active DOFs will be checked
+    @param trajectory: the trajectory whose start configuration will be checked
+    @returns: True if the robot's active DOFs match the given trajectory
+              False if one or more active DOFs differ by DOF resolution
+    """
+    # Get current active configuration of robot.
+    dof_indices = robot.GetActiveDOFIndices()
+    dof_values = robot.GetActiveDOFValues()
+    dof_resolutions = robot.GetActiveDOFResolutions()
```
mkoval: The DOFs in the trajectory may not be (and often are not) active during trajectory execution. This function should check the DOFs included in `traj.GetConfigurationSpecification()`, not the active DOFs.

psigen: Is this the correct logic?  If the DOFs are not active, we are not moving them.  

So why does it matter where they are *only* for the start?  If we have to check the start locations of all the non-active DOFs in the trajectory, don't we have to check that those DOFs are always at that location at every time over the whole trajectory?  It seems like an all or nothing thing.

src/prpy/base/robot.py
psigen: Addressed in 798daac9b92578e2e7a0b24edd77de59f05c9565.

src/prpy/util.py
mkoval: I am nearly 100% certain that `SetPath` ignores the active DOFs.

psigen: Ok, but what *should* it do?  Keep in mind that we are under no obligation to use OpenRAVE defaults here.  ExecuteTrajectory can splice the trajectory however we see fit before passing it into the SetPath function.

It seems like completely ignoring ActiveDOFs is not particularly intuitive behavior.  Doesn't it?

mkoval: I still think that `ExecuteTrajectory` should ignore the active DOFs. Here are three concrete reasons:

1. **Backwards Compatibility.** `ExecuteTrajectory` has always ignored the active DOFs.
2. **OpenRAVE Convention.** `ExecuteTrajectory` performs the same function as `SetPath`, which ignores active DOFs.
3. **Safety.** The trajectory formed by an arbitrary subset of the trajectory's DOFs may not be collision free.

Honoring the active DOFs also leads to counterintuitive behavior when mixed with robot state savers. Depending upon the active DOFs the following code could: (1) execute a right arm trajectory, (2) execute a trajectory on a subset of the right arm's DOFs, or (3) do nothing. See below:

```python
traj = robot.right_arm.PlanToNamedConfiguration('home')
robot.ExecutePath(traj)
```

Finally, I think active DOFs and manipulators are a generally bad idea. The "active" concept introduce additional state that is a constant headache to manage. I'd like to entirely move PrPy away from active DOFs in the future. We're already considering doing so in the planning pipeline, because the concept of an "active manipulator" breaks on bimanual tasks.

cdellin: Note also that the documentation for OpenRAVE [0] suggests that the notion of active DOFs is meant only for planners (not executors, etc):

> Set the joint indices and affine transformation dofs that the planner should use.

[0] http://openrave.org/docs/latest_stable/coreapihtml/classOpenRAVE_1_1RobotBase.html#a818d340885e3a5b0372c7bae025430af

psigen: Ok, these are compelling arguments, thank guys!

mkoval:
LGTM with the latest changes.

PR #148, opened 2015-06-16
Changed GetDuration=0.0 error to warning.

psigen:
Currently, there are multiple non-compliant algorithms in the post-processing pipeline that generate multiple-waypoint zero-duration trajectories if given a single-waypoint trajectory.

We will need to keep this as a warning until they are updated.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -459,7 +459,10 @@ class Robot(openravepy.Robot):
 
         # Verify that the trajectory is timed.
         if traj.GetDuration() <= 0.0:
-            raise ValueError('Attempted to execute untimed trajectory.')
+            import warnings
+            warnings.warn('Executing zero-length trajectory. Please update the'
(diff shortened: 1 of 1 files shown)


src/prpy/base/robot.py
```diff
@@ -459,7 +459,10 @@ def ExecuteTrajectory(self, traj, defer=False, timeout=None, period=0.01, **kwar
 
         # Verify that the trajectory is timed.
         if traj.GetDuration() <= 0.0:
```
mkoval: This check conflates two types of errors:

1. untimed trajectory that is missing the `deltatime` group
2. timed trajectory with zero duration

These have very different causes, so they should produce different error messages. I am fine with leaving (2) as a warning, for now, until we squash the remaining bugs in the planning pipeline. Can you add a check for (1) that throws an exception?

psigen: Addressed in c616fa1.

mkoval:
:+1:

PR #149, opened 2015-06-16
Bugfix: SimplifyTrajectory calls in retimer

psigen:
SimplifyTrajectory is very simple and can't safely reinterpolate timed trajectories.  This patch modifies the retimers to not call SimplifyTrajectory on already timed trajectories, since it will just throw an error and fail anyway.


mkoval:
LGTM.

PR #152, opened 2015-06-23
Added tsr conversion methods for JSON and YAML.

psigen:
This fixes the naming convention on TSR serialization to use `to_`/`from_` and adds convenience conversion methods for JSON and YAML.

### DIFF
--- a/src/prpy/planning/cbirrt.py
+++ b/src/prpy/planning/cbirrt.py
@@ -245,7 +245,7 @@ class CBiRRTPlanner(BasePlanner):
 
         if tsr_chains is not None:
             for tsr_chain in tsr_chains:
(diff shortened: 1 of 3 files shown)


mkoval:
There are several references to `serialize_dict` and `deserialize_dict` in `serialization.py`. Can you update those to use the new names?

It's also worth grepping the directory for their use to make sure you didn't miss anything.

psigen:
Ok, refactored other referenced in prpy to use `to/from_dict`.

I also moved the CBiRRT TSR serialize functions into `cbirrt.py`, because nothing in the universe should be using them.  Nothing in prpy should even know they ever existed.

@mkoval: can you test that this still works for you?

psigen:
@mkoval, does this address your concerns?

mkoval:
:+1:

PR #156, opened 2015-07-10
Feature/generic object tsr

rachelholladay:
This add the functionary of removing the requirement that tsrs be tied to specific objects. This allows for object-generic tsrs, which is important for some of the coming gesture tsrs. 

There is also a slight edit to the base code that @jeking04 would have to vouch for as I'm not totally sure of its status.


src/prpy/base/mobilebase.py
```diff
@@ -85,20 +85,17 @@ def Forward(self, meters, execute=True, direction=None, **kw_args):
...
+            offset_pose = numpy.eye(4)
+            offset_pose[0:3, 3] = meters * direction
+            goal_pose = numpy.dot(start_pose, offset_pose)
+            
+        path = create_affine_trajectory(self.robot, [ start_pose, goal_pose ])
+        if execute:
+            return self.robot.ExecutePath(path, **kw_args)
         else:
-            raise NotImplementedError('DriveForward is not implemented')
+            return path
```
mkoval: I'm not up-to-date on the recent changes to `MobileBase`. I agree with whatever @jeking04 says. :smile:

src/prpy/tsr/tsrlibrary.py
```diff
@@ -80,18 +81,28 @@ def __call__(self, kinbody, action_name, *args, **kw_args):
...
             logger.debug('Inferred KinBody name "%s" for TSR.', kinbody_name)
 
+        f = None
         try:
             f = self.all_factories[self.robot_name][kinbody_name][action_name]
-        except KeyError:
-            raise KeyError('There is no TSR factory registered for action "{:s}"'
-                           ' with robot "{:s}" and object "{:s}".'.format(
-                           action_name, self.robot_name, kinbody_name))
+        except KeyError, ignored:
```
mkoval: I believe the `except Exception, name:` syntax is deprecated. Please use `except KeyError as ignored:` instead. Or just `except KeyError:` if you don't plan to use the variable at all.

mkoval:
LGTM other than those two comments. I'll merge after I hear back from @jeking04 on the mobile base changes.

src/prpy/base/mobilebase.py
jeking04: I think this is something we want to keep.  

If I recall, previously we were raising the error in prpy because we might not always be capable of executing these trajectories on the real robot. Then in herbpy, we were overriding this function to actually return the path, because we know with herb we can execute them.

This flips that thinking. Basically, we assume all robots can execute these affine trajectories. If a robot can't, it needs to override this function.  I think this also matches the thinking of the regular planning pipeline.

mkoval:
As always, I agree with @jeking04.

PR #157, opened 2015-07-14
Add support for numpy <1.8 which don't support norm(axis=1)

cdellin:
Numpy 1.8, released in October 2013, added an optional `axis` argument to `numpy.linalg.norm()`, which is used in prpy by `ik_ranking.py`.  My system uses an older version of Numpy, and I'd like it to be supported.  This PR does not change the code itself, but instead defines a custom function only when necessary which handles the axis case using the supported `numpy.apply_along_axis()` function.

### DIFF
--- a/src/prpy/__init__.py
+++ b/src/prpy/__init__.py
@@ -32,3 +32,4 @@ import base, dependency_manager, logger, ik_ranking, planning, simulation, tsr, 
 from named_config import ConfigurationLibrary
 from clone import Clone, Cloned
 from bind import bind_subclass
+import compatibility
--- a/src/prpy/compatibility.py
+++ b/src/prpy/compatibility.py
@@ -1,0 +1,14 @@
+from distutils.version import LooseVersion
+import numpy
+
+# handle older versions of numpy without norm axis argument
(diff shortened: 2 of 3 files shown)


mkoval:
The change looks good.

However, part of me is inclined to say that you should upgrade NumPy. :smile:

cdellin:
I am sympathetic to the suggestion that I upgrade Numpy (-: it'll just probably wait until after this whole PhD thing is done so I can upgrade to Jessie all at once.

psigen:
Is this going to come up in other places?  Maybe it makes more sense to load a monkeypatch file like `prpy.compatibility.numpy` from  `prpy/__init__.py` that does this, so we can put together all the various numpy compatibility hacks one might need (e.g. `numpy.allclose`)

mkoval:
:+1: Could you move this to a dedicated file?

cdellin:
Done. For now I just added a file called compatibility.py.  If we end up with more of these, we can split it up into a compatibility module (although we can't name a file therein numpy.py, python gets very confused about that).

psigen:
:+1: seems good to me.

mkoval:
@jeking04 confirmed that this doesn't break anything on 14.04. (Thanks @jeking04!)

PR #159, opened 2015-07-19
Feature: TSR and TSRs cleanup and helper functions

siddhss5:
This PR cleans up some of the existing TSR and TSRs functions, and adds new helper functions for computing distance to TSR and TSRs, among others.

### DIFF
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -336,24 +336,24 @@ class TSRChain(object):
         x_dict = yaml.safe_load(x, *args, **kw_args)
         return TSR.from_dict(x_dict)
 
-    def is_valid(self, xyzrpy, ignoreNAN=False):
+    def is_valid(self, xyzrpy_list, ignoreNAN=False):
         """
         Checks if a xyzrpy list is a valid sample from the TSR.
-        @param xyzrpy a list of xyzrpy values
+        @param xyzrpy_list a list of xyzrpy values
         @param ignoreNAN (optional, defaults to False) ignore NaN xyzrpy
         @return a list of 6x1 vector of True if bound is valid and False if not
         """
 
-        if len(xyzrpy) != len(self.TSRs):
+        if len(xyzrpy_list) != len(self.TSRs):
             raise('Sample must be of equal length to TSR chain!')
 
         check = []
         for idx in range(len(self.TSRs)):
-            check.append(self.TSRs[idx].is_valid(xyzrpy[idx], ignoreNAN))
+            check.append(self.TSRs[idx].is_valid(xyzrpy_list[idx], ignoreNAN))
 
         return check
 
-    def to_transform(self, xyzrpy):
+    def to_transform(self, xyzrpy_list):
         """
         Converts a xyzrpy list into an
         end-effector transform.
@@ -361,47 +361,47 @@ class TSRChain(object):
         @param  a list of xyzrpy values
         @return trans 4x4 transform
         """
-        check = self.is_valid(xyzrpy)
+        check = self.is_valid(xyzrpy_list)
         for idx in range(len(self.TSRs)):
             if not all(check[idx]):
-                raise ValueError('Invalid xyzrpy', check)
-        T0_w = self.TSRs[0].T0_w
+                raise ValueError('Invalid xyzrpy_list', check)
+        T_sofar = self.TSRs[0].T0_w
         for idx in range(len(self.TSRs)):
             tsr_current = self.TSRs[idx]
-            tsr_current.T0_w = T0_w
-            T0_w = tsr_current.to_transform(xyzrpy[idx])
+            tsr_current.T0_w = T_sofar
+            T_sofar = tsr_current.to_transform(xyzrpy_list[idx])
 
-        return T0_w
+        return T_sofar
 
-    def sample_xyzrpy(self, xyzrpy=None):
+    def sample_xyzrpy(self, xyzrpy_list=None):
         """
-        Samples from Bw to generate an xyzrpy sample
+        Samples from Bw to generate a list of xyzrpy samples
         Can specify some values optionally as NaN.
 
-        @param xyzrpy   (optional) a list of Bw with float('nan') for
+        @param xyzrpy_list   (optional) a list of Bw with float('nan') for
                         dimensions to sample uniformly.
         @return sample  a list of sampled xyzrpy
         """
 
-        if xyzrpy is None:
-            xyzrpy = [NANBW]*len(self.TSRs)
+        if xyzrpy_list is None:
+            xyzrpy_list = [NANBW]*len(self.TSRs)
 
         sample = []
         for idx in range(len(self.TSRs)):
-            sample.append(self.TSRs[idx].sample_xyzrpy(xyzrpy[idx]))
+            sample.append(self.TSRs[idx].sample_xyzrpy(xyzrpy_list[idx]))
 
         return sample
 
-    def sample(self, xyzrpy=None):
+    def sample(self, xyzrpy_list=None):
         """
         Samples from the Bw chain to generate an end-effector transform.
         Can specify some Bw values optionally.
 
-        @param xyzrpy   (optional) a list of xyzrpy with float('nan') for
-                        dimensions to sample uniformly.
-        @return T0_w    4x4 transform
+        @param xyzrpy_list   (optional) a list of xyzrpy with float('nan') for
+                             dimensions to sample uniformly.
+        @return T0_w         4x4 transform
         """
-        return self.to_transform(self.sample_xyzrpy(xyzrpy))
+        return self.to_transform(self.sample_xyzrpy(xyzrpy_list))
 
     def distance(self, trans):
         """
@@ -412,10 +412,10 @@ class TSRChain(object):
         """
         import scipy.optimize
 
-        def objective(xyzrpy):
-            bw_stack = xyzrpy.reshape(len(self.TSRs), 6)
-            bwtrans = self.to_transform(bw_stack)
-            return prpy.util.GeodesicDistance(bwtrans, trans)
+        def objective(xyzrpy_list):
+            xyzrpy_stack = xyzrpy_list.reshape(len(self.TSRs), 6)
+            tsr_trans = self.to_transform(xyzrpy_stack)
+            return prpy.util.GeodesicDistance(tsr_trans, trans)
 
         bwinit = []
         bwbounds = []
@@ -443,7 +443,7 @@ class TSRChain(object):
         """
         Converts an end-effector transform to a list of xyzrpy values
         @param  trans  4x4 transform
-        @return xyzrpy list of xyzrpy values
+        @return xyzrpy_list list of xyzrpy values
         """
-        _, xyzrpy = self.distance(trans)
-        return xyzrpy
+        _, xyzrpy_list = self.distance(trans)
+        return xyzrpy_list
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -35,6 +35,8 @@ import numpy
 import numpy.random
 import kin
 
+NANBW = numpy.ones(6)*float('nan')
+
 """
 Functions for Serializing TSRs and TSR Chains
 
@@ -153,7 +155,7 @@ class TSR(object):  # force new-style class
                  for i, x in enumerate(Bwvals)]
         return all(check)
 
-    def sample(self, vals=numpy.ones(6)*float('nan')):
+    def sample(self, vals=NANBW):
         """
         Samples from Bw to generate an end-effector transform.
         Can specify some Bw values optionally.
@@ -287,7 +289,7 @@ class TSRChain(object):
         if len(self.TSRs) == 0:
             return None
         if vals is None:
-            vals = [numpy.ones(6)*float('nan')]*len(self.TSRs)
+            vals = NANBW*len(self.TSRs)
 
         T0_w = self.TSRs[0].T0_w
         for idx in range(len(self.TSRs)):
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -76,7 +76,7 @@ class TSR(object):
         """
         Converts a rotation matrix to one valid rpy
         @param rot 3x3 rotation matrix
-        @return rpy 3x1 rpy
+        @return rpy (3,) rpy
         """
         rpy = numpy.zeros(3)
         if not (abs(abs(rot[2, 0]) - 1) < EPSILON):
@@ -114,7 +114,7 @@ class TSR(object):
     def rpy_to_rot(rpy):
         """
         Converts an rpy to a rotation matrix
-        @param rpy 3x1 rpy
+        @param rpy (3,) rpy
         @return rot 3x3 rotation matrix
         """
         rot = numpy.zeros((3, 3))
@@ -154,9 +154,9 @@ class TSR(object):
         """
         Checks whether an xyz value is within a given xyz bounds.
         Main issue: dealing with roundoff issues for zero bounds
-        @param xyz a 3x1 xyz value
+        @param xyz a (3,) xyz value
         @param Bw bounds on xyz
-        @return check a 3x1 vector of True if within and False if outside
+        @return check a (3,) vector of True if within and False if outside
         """
         # Check bounds condition on XYZ component.
         xyzcheck = [((x + EPSILON) >= Bw[i, 0]) and
@@ -171,9 +171,9 @@ class TSR(object):
         Assumes all values in the bounds are [-pi, pi]
         Two main issues: dealing with roundoff issues for zero bounds and
         Wraparound for rpy.
-        @param rpy a 3x1 rpy value
+        @param rpy a (3,) rpy value
         @param Bw bounds on rpy
-        @return check a 3x1 vector of True if within and False if outside
+        @return check a (3,) vector of True if within and False if outside
         """
         # Unwrap rpy to [-pi, pi].
         rpy = (numpy.array(rpy) + pi) % (2*pi) - pi
@@ -202,7 +202,7 @@ class TSR(object):
         Based on: http://staff.city.ac.uk/~sbbh653/publications/euler.pdf
         @param rot 3x3 rotation matrix
         @param Bw bounds on rpy
-        @return check a 3x1 vector of True if within and False if outside
+        @return check a (3,) vector of True if within and False if outside
         @return rpy the rpy consistent with the bound or None if nothing is
         """
         if not (abs(abs(rot[2, 0]) - 1) < EPSILON):
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -85,12 +85,13 @@ class TSR(object):
                   ypr[2], ypr[1], ypr[0]]
         return xyzrpy
 
-    def is_valid(self, xyzrpy):
+    def is_valid(self, xyzrpy, ignoreNAN=False):
         """
         Checks if a xyzrpy is a valid sample from the TSR.
         Two main issues: dealing with roundoff issues for zero bounds and
         Wraparound for rpy.
         @param xyzrpy 6x1 vector of Bw values
+        @param ignoreNAN (optional, defaults to False) ignore NaN xyzrpy
         @return True if valid and False if not
         """
         from math import pi
@@ -99,11 +100,21 @@ class TSR(object):
         xyzcheck = [((x >= Bw_xyz[i, 0]) and (x <= Bw_xyz[i, 1]))
                     or (Bw_xyz[i, 1] - Bw_xyz[i, 0] < EPSILON)
                     for i, x in enumerate(xyzrpy[0:3])]
+
+        # Unwrap all rotations to [-pi, pi]
         Bw_rpy = (self.Bw[4:6, :] + pi) % (2*pi) - pi
+        rpy = (xyzrpy[4:6] + pi) % (2*pi) - pi
         rpycheck = [((x >= Bw_rpy[i, 0]) and (x <= Bw_rpy[i, 1]))
                     or (Bw_rpy[i, 1] - Bw_rpy[i, 0] < EPSILON)
-                    for i, x in enumerate(xyzrpy[3:6])]
-        return all(xyzcheck) and all(rpycheck)
+                    for i, x in enumerate(rpy)]
+
+        check = numpy.hstack(xyzcheck, rpycheck)
+
+        # Strip out Nans if needed
+        if ignoreNAN is True:
+            check = check[~numpy.isnan(xyzrpy)]
+
+        return all(check)
 
     def contains(self, trans):
         """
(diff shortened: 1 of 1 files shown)


src/prpy/tsr/tsr.py
```diff
@@ -52,36 +61,132 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+        Wraparound for rpy.
+        @param xyzrpy 6x1 vector of Bw values
+        @param ignoreNAN (optional, defaults to False) ignore NaN xyzrpy
+        @return True if valid and False if not
+        """
+        Bw_xyz = self.Bw[0:3, :]
+
+        xyzcheck = [(((x + EPSILON) >= Bw_xyz[i, 0]) and 
+                     ((x - EPSILON) <= Bw_xyz[i, 1]))
+                    or (Bw_xyz[i, 1] - Bw_xyz[i, 0] < EPSILON)
```
mkoval: Why is this extra check for a point TSR necessary?

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +61,132 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+                    for i, x in enumerate(xyzrpy[0:3])]
+
+        # Unwrap all rotations to [-pi, pi]
+        Bw_rpy = (self.Bw[3:6, :] + pi) % (2*pi) - pi
+        rpy = numpy.add(xyzrpy[3:6], pi) % (2*pi) - pi
+        rpycheck = [(((x + EPSILON) >= Bw_rpy[i, 0]) and 
+                     ((x - EPSILON) <= Bw_rpy[i, 1]))
+                    or (Bw_rpy[i, 1] - Bw_rpy[i, 0] < EPSILON)
+                    for i, x in enumerate(rpy)]
```
mkoval: This is technically correct, but I'm not sure if it's the logic that we want. This makes it impossible to create a TSR that includes `pi`, but not `0`. One alternative is to treat the range as signed, so `[ -0.5, 0.5 ]` refers to the range that includes the origin and `[ -0.5, 0.5 ]` refers to the range that includes `pi`.

I'm not sure if this is better or worse. Thoughts?

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +61,132 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+        return all(check)
+
+    def contains(self, trans):
+        """
+        Checks if the TSR contains the transform
+        @param  trans 4x4 transform
+        @return       True if inside and False if not
+        """
+        xyzrpy = self.to_xyzrpy(trans)
+        return self.is_valid(xyzrpy)
```
mkoval: Instead of a `bool`, it might be nice to return the mask. This could help us write very informative error messages about why a constraint is being violated.

src/prpy/tsr/tsr.py
```diff
@@ -122,7 +227,7 @@ def from_json(x, *args, **kw_args):
     def to_yaml(self):
         """ Convert this TSR to a YAML string. """
         import yaml
-        return yaml.dumps(self.to_dict())
+        return yaml.dump(self.to_dict())
```
mkoval: I think we want `dumps` here, since `dump` expects a `File`-like object. What was the motivation behind this change?

src/prpy/tsr/tsr.py
```diff
@@ -231,14 +336,113 @@ def from_yaml(x, *args, **kw_args):
...
             tsr_current.T0_w = T0_w
-            T0_w = tsr_current.sample()
+            T0_w = tsr_current.to_transform(xyzrpy[idx])
 
         return T0_w
+
+    def sample_xyzrpy(self, xyzrpy=None):
+        """
+        Samples from Bw to generate an xyzrpy sample
+        Can specify some values optionally as NaN.
```
mkoval: This comment is inaccurate. This returns a list of `xyzrpy` samples, where the `i`-th entry corresponds to the values for the `i`-th TSR in the chain.

src/prpy/tsr/tsr.py
```diff
@@ -231,14 +336,113 @@ def from_yaml(x, *args, **kw_args):
...
+        if not self.is_valid(xyzrpy):
+            raise ValueError('Invalid xyzrpy')
         T0_w = self.TSRs[0].T0_w
         for idx in range(len(self.TSRs)):
             tsr_current = self.TSRs[idx]
             tsr_current.T0_w = T0_w
-            T0_w = tsr_current.sample()
+            T0_w = tsr_current.to_transform(xyzrpy[idx])
 
         return T0_w
```
mkoval: Shouldn't there be some recursive multiplication in here to compose the relative transforms?

src/prpy/tsr/tsr.py
siddhss5: This was a Rachel thing. I don't really know anything about the serialization bits.

rachelholladay: @mkoval When I checked, `dumps` wasn't a command. I naively guessed @siddhss5 typo-ed and meant `dump`.

psigen: > *yaml.dump* accepts the second optional argument, which must be an open
text or binary file. In this case,*yaml.dump* will write the produced YAML
document into the file. Otherwise, *yaml.dump* returns the produced
document.

src/prpy/tsr/tsr.py
siddhss5: Fixed in https://github.com/personalrobotics/prpy/commit/4052dd7647f53831cad78b0374d53dae57b41b0e.

src/prpy/tsr/tsr.py
siddhss5: Hmm, could you take a stab at proposing some Pythonic logic for this @mkoval? I see your point and it makes a lot of sense.

src/prpy/tsr/tsr.py
mkoval: :+1: Interesting. I wonder why the equivalent for JSON is `json.dumps`.

src/prpy/tsr/tsr.py
```diff
@@ -41,6 +48,8 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
             Tw_e = numpy.eye(4)
         if Bw is None:
             Bw = numpy.zeros((6, 2))
+        elif any((Bw[:, 1]-Bw[:, 0]) > 2*pi + EPSILON):
+            raise(ValueError('Bw range must be within 2*PI'))
```
cdellin: Two quick questions about "what it means to be a TSR":

1. Must all W transforms contain the identity pose (i.e. must the first column of Bw be all-nonpositive, and the second be all-nonnegative)?  There's no fundamental reason why this must be so, but IIRC CBiRRT does not like bounds on W that exclude identity.

2. Dmitry's TSR code [0] claims that Bw[:,0] can be greater than Bw[:,1], signifying "an axis flip (not an error)".  Does this code support this?  If so, this check should have an absolute value in it.

In any case, it might be a good idea to write up a small section on the prpy README describing what a TSR is.

[0] https://github.com/personalrobotics/comps/blob/master/cbirrt2/TaskSpaceRegion.cpp#L65

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +61,130 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+        """
+        if len(xyzrpy) != 6:
+            raise ValueError('vals must be of length 6')
+        if not self.is_valid(xyzrpy):
+            raise ValueError('Invalid xyzrpy')
+
+        xyzypr = [xyzrpy[0], xyzrpy[1], xyzrpy[2],
+                  xyzrpy[5], xyzrpy[4], xyzrpy[3]]
+        Tw = kin.pose_to_H(kin.pose_from_xyzypr(xyzypr))
         trans = numpy.dot(numpy.dot(self.T0_w, Tw), self.Tw_e)
```
cdellin: How do people feel about `trans = reduce(numpy.dot, [self.T0_w, Tw, self.Tw_e])` for things like this?

src/prpy/tsr/tsr.py
cdellin: I think `contains()` sounds to me like it should return a `bool`, but I agree it would be nice to have what Mike wants in another method ... perhaps `get_violations()` or something?

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +61,130 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+            return prpy.util.GeodesicDistance(bwtrans, trans)
+
+        bwinit = (self.Bw[:, 0] + self.Bw[:, 1])/2
+        bwbounds = [(self.Bw[i, 0], self.Bw[i, 1]) for i in range(6)]
+
+        bwopt, dist, info = scipy.optimize.fmin_l_bfgs_b(
+                                objective, bwinit, fprime=None,
+                                args=(),
+                                bounds=bwbounds, approx_grad=True)
+        return dist, bwopt
```
cdellin: I suppose if we were to use this often, we could at least compute the gradient analytically, right?

src/prpy/tsr/tsr.py
```diff
@@ -231,14 +334,113 @@ def from_yaml(x, *args, **kw_args):
...
+        """
+
+        if len(xyzrpy) != len(self.TSRs):
+            raise('Sample must be of equal length to TSR chain!')
+
+        for idx in range(len(self.TSRs)):
+            if not self.TSRs[idx].is_valid(xyzrpy[idx], ignoreNAN):
+                return False
+
+        return True
```
cdellin: The parameter `xyzrpy` should be `xyzrpy_list` or something, and I would rewrite the loop using `zip()`.

src/prpy/tsr/tsr.py
cdellin: @mkoval, no overwriting each TSR's T0_w does that.

I would call the variable `T0_w` something different here though, perhaps `T_sofar` or something?

src/prpy/tsr/tsr.py
```diff
@@ -231,14 +334,113 @@ def from_yaml(x, *args, **kw_args):
...
+        for idx in range(len(self.TSRs)):
+            Bw = self.TSRs[idx].Bw
+            bwinit.extend((Bw[:, 0] + Bw[:, 1])/2)
+            bwbounds.extend([(Bw[i, 0], Bw[i, 1]) for i in range(6)])
+
+        bwopt, dist, info = scipy.optimize.fmin_l_bfgs_b(
+                                objective, bwinit, fprime=None,
+                                args=(),
+                                bounds=bwbounds, approx_grad=True)
+        return dist, bwopt.reshape(len(self.TSRs), 6)
```
cdellin: This looks good, I would just rename `xyzrpy` as `xyzrpy_stacked` or something like that.

src/prpy/tsr/tsr.py
siddhss5: I saw this too late. I like @cdellin's idea of `get_violations()` but I just committed the mask for `contains()` here https://github.com/personalrobotics/prpy/commit/b232cc329da469688593c5d4defd54a0ba439634.

siddhss5:
At this point, I have don't have any more time to implement the enhancements above. I do like all of them, so I'm handing off the baton to anyone else who'd like to implement them.

src/prpy/tsr/tsr.py
siddhss5: This goes back to @mkoval's previous comment on the expressiveness of the TSR representation. I'm quite ambivalent to any of this, so please implement whatever seems reasonable.

src/prpy/tsr/tsr.py
siddhss5: Oooo so pretty :clap:.

mkoval:
@cdellin, can you seize the baton? :smile:

src/prpy/tsr/tsr.py
mkoval: I see. This makes sense.

src/prpy/tsr/tsr.py
siddhss5: The advantage of explicitly looping is the small gain in efficiency and the perhaps larger gain in code clarity? :)

siddhss5:
I'll keep working on it myself. I'm pretty close to being done. The main thing to figure out is the wraparound pi normalization, which I should be able to at some point.

src/prpy/tsr/tsr.py
siddhss5: No idea how to do this efficiently. Feel free to add this yourself or send me some math/code that does it.

src/prpy/tsr/tsr.py
siddhss5: I'm marking this as resolved, right? I have no JSON skillz.

src/prpy/tsr/tsr.py
siddhss5: This is addressed in https://github.com/personalrobotics/prpy/commit/5829a37ced1aed314f1da9d4ad7ca6b3e52cb517.

src/prpy/tsr/tsr.py
siddhss5: This is addressed in https://github.com/personalrobotics/prpy/commit/5829a37ced1aed314f1da9d4ad7ca6b3e52cb517.

src/prpy/tsr/tsr.py
siddhss5: Variable renaming for clarity addressed in https://github.com/personalrobotics/prpy/commit/5829a37ced1aed314f1da9d4ad7ca6b3e52cb517.

src/prpy/tsr/tsr.py
siddhss5: This is addressed in https://github.com/personalrobotics/prpy/commit/5829a37ced1aed314f1da9d4ad7ca6b3e52cb517.

src/prpy/tsr/tsr.py
siddhss5: This is addressed in https://github.com/personalrobotics/prpy/commit/576014b6388f12451bba9927390d2503d4086928.

src/prpy/tsr/tsr.py
siddhss5: So, if a bound straddles pi, can't I just transform `T0_w` about that axis by pi and rotate the bounds by pi so that they straddle 0?

siddhss5: @psigen suggested a simple and elegant solution [which @mkoval might also have alluded to in his earlier post]: treat the bounds as signed, i.e. as `from -> to` instead of `[min, max]` and modify `Bw[i,1]` to `Bw[i,1] + 2pi` if `Bw[i,1] < Bw[i,0]`. This will shift it out of the `pi` wraparound. Of course, this means, `Bw[i,1]` will no longer be bounded `[-pi,pi]` but can be as high as `pi+2pi=3pi` for example if we want the almost great circle `[pi, pi-eps]`. But I'm willing to live with that.

src/prpy/tsr/tsr.py
```diff
@@ -41,9 +46,23 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+
         self.T0_w = T0_w
         self.Tw_e = Tw_e
         self.Bw = Bw
+
+        # We will now create a continuous version of the bound to maintain:
+        # 1. Bw[i,1] > Bw[i,0] which is necessary for LBFGS-B
+        # 2. signed rotations, necessary for expressiveness
+        Bw_cont = numpy.copy(Bw)
+        Bw_cont[3:6, :] = (Bw_cont[3:6, :] + pi) % (2*pi) - pi
```
mkoval: Will this normalization always work? If `Bw_cont[3:6, :] < -pi`, then adding `pi` will not make the value positive. I think this should be:
```python
k = max(int(-Bw_cont[3:6, :].min() / pi), 0) + 1
Bw_cont[3:6, :] = (Bw_cont[3:6, :] + k * pi) % (2*pi) - pi
```
I always screw up angle normalization, so I very well might be wrong here. :sweat:

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+
+    @staticmethod
+    def trans_to_xyzrpy(trans):
+        """
+        Converts a transformation matrix to one valid xyzrpy
+        @param trans 4x4 transformation matrix
+        @return xyzrpy 6x1 xyzrpy
+        """
+        xyz, rot = trans[0:3, 3], trans[0:3, 0:3]
+        rpy = TSR.rot_to_rpy(rot)
```
mkoval: I'd prefer to make this a `@classmethod` and replace `TSR.rot_to_rpy` with `cls.rot_to_rpy`. It's nice to DRY if we ever have to rename the class.

We'll probably never rename `TSR`, so this is just a nitpick. Feel free to disregard.

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+        """
+        Converts an xyzrpy to a transformation matrix
+        @param xyzrpy 6x1 xyzrpy vector
+        @return trans 4x4 transformation matrix
+        """
+        trans = numpy.zeros((4, 4))
+        trans[3][3] = 1.0
+        xyz, rpy = xyzrpy[0:3], xyzrpy[3:6]
+        trans[0:3, 3] = xyz
+        rot = TSR.rpy_to_rot(rpy)
```
mkoval: Same comment here about `@classmethod`.

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+        Checks whether an rpy value is within a given rpy bounds.
+        Assumes all values in the bounds are [-pi, pi]
+        Two main issues: dealing with roundoff issues for zero bounds and
+        Wraparound for rpy.
+        @param rpy a 3x1 rpy value
+        @param Bw bounds on rpy
+        @return check a 3x1 vector of True if within and False if outside
+        """
+        # Unwrap rpy to [-pi, pi].
+        rpy = (numpy.array(rpy) + pi) % (2*pi) - pi
```
mkoval: Same comment as above. Will this always work?

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+    def rpy_within_bounds(rpy, Bw):
+        """
+        Checks whether an rpy value is within a given rpy bounds.
+        Assumes all values in the bounds are [-pi, pi]
+        Two main issues: dealing with roundoff issues for zero bounds and
+        Wraparound for rpy.
+        @param rpy a 3x1 rpy value
+        @param Bw bounds on rpy
+        @return check a 3x1 vector of True if within and False if outside
+        """
```
mkoval: This method takes a NumPy array with shape `(3,)`, not `(3,1)`. NumPy differentiates between a three-element one-dimensional array and a (3 x 1) two-dimensional array, so I think it is misleading to say `a 3x1 vector` in the docstring.

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+            # Not a singularity. Two pitch solutions
+            psol = -numpy.arcsin(rot[2, 0])
+            for p in [psol, (pi - psol)]:
+                rpy = numpy.zeros(3)
+                rpy[0] = numpy.arctan2((rot[2, 1]/numpy.cos(p)),
+                                       (rot[2, 2]/numpy.cos(p)))
+                rpy[1] = p
+                rpy[2] = numpy.arctan2((rot[1, 0]/numpy.cos(p)),
+                                       (rot[0, 0]/numpy.cos(p)))
+                rpycheck = TSR.rpy_within_bounds(rpy, Bw)
```
mkoval: Ditto about `@classmethod`.

src/prpy/tsr/tsr.py
```diff
@@ -52,36 +71,323 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
+                # check the four r-y Bw corners
+                rpy_list = []
+                rpy_list.append([Bw[2, 0] + r_offset, pi/2, Bw[2, 0]])
+                rpy_list.append([Bw[2, 1] + r_offset, pi/2, Bw[2, 1]])
+                rpy_list.append([Bw[0, 0], pi/2, Bw[0, 0] - r_offset])
+                rpy_list.append([Bw[0, 1], pi/2, Bw[0, 1] - r_offset])
+                for rpy in rpy_list:
+                    rpycheck = TSR.rpy_within_bounds(rpy, Bw)
+                    # No point checking anything if pi/2 not in Bw
+                    if (rpycheck[1] is False):
```
mkoval: These parens are un-Pythonic. :smile:

src/prpy/tsr/tsr.py
mkoval: @cdellin Did you derive the gradient yet? :smile:

src/prpy/tsr/tsr.py
mkoval: Yeah, this looks good.

mkoval:
Overall, this looks good. I will merge as soon as we confirm that the angle normalization code is correct.

I also made a few nitpicks about syntax, but they're all minor.

src/prpy/tsr/tsr.py
siddhss5: Err, what? Behold:
```
import matplotlib.pyplot as plt
import numpy
import numpy.pi as pi
x = numpy.linspace(-10*pi, 10*pi, 1000)
y = [((i + pi) % (2*pi) - pi) for i in x]
plt.plot(x, y)
plt.show()
```

src/prpy/tsr/tsr.py
siddhss5: Disregarding :).

src/prpy/tsr/tsr.py
siddhss5: Ya.

src/prpy/tsr/tsr.py
siddhss5: Clarified in https://github.com/personalrobotics/prpy/commit/e49f68ab676c252f9e8340e98a4c9af7e063c9eb.

src/prpy/tsr/tsr.py
mkoval: I don't know why this works. But the graph looks right, so I'll just accept it as truth.

siddhss5:
WRT @cdellin's comment, I agree writing robust conversion code is hard. Also agree with @psigen that we should create a new PR to decide where some of the conversion code should go. My preference is `util.py` but I'm happy to consider other alternatives. I'm also annoyed by `kin.py` as it replicates some of the existing conversion functionality in `OpenRAVE`.

src/prpy/tsr/tsr.py
siddhss5: I hate you.

cdellin:
:thumbsup:

mkoval:
I created #165 to track the conversion function cleanup. There are no major complaints left, so I am merging this.

PR #161, opened 2015-08-02
changed acceleration limits

Stefanos19:
ADA moves slightly faster after the Kinova firmware upgrade, while still having a high success-rate in grasping the bites in the ada-meal scenario demo.

### DIFF
--- a/src/prpy/base/mico.py
+++ b/src/prpy/base/mico.py
@@ -44,7 +44,7 @@ class Mico(Manipulator):
 
         with env:
             accel_limits = robot.GetDOFAccelerationLimits()
-            accel_limits[self.GetIndices()] = 1.0
+            accel_limits[self.GetIndices()] = 1.75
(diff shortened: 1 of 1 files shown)


mkoval:
Please verify these in a controlled test before merging this; i.e. command the arm to go from zero velocity to maximum velocity instantaneously and see how fast it can accelerate.

Stefanos19:
Ok sounds good.

Stefanos19:
So, I verified the acceleration limits. I will set them to these values and make a new PR. 

j1
Acceleration limit: 1.4507

j2
Acceleration limit: 1.5633

j3
Acceleartion limit: 1.5596

j4
Acceleration limit; 1.4955

j5
Acceleartion limit: 1.4807

j6
Acceleration Limit: 1.4887

mkoval:
LGTM. Thanks @Stefanos19!

PR #163, opened 2015-08-04
Created internal continuous range `Bw_cont`.

psigen:
This separates out the continuous and non-continuous Bw values so the public API for a TSR can strictly deal with [-pi, pi]

### DIFF
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -123,6 +123,10 @@ class TSR(object):
                     for i, x in enumerate(xyzrpy[0:3])]
 
         Bw_rpy = self.Bw[3:6, :]
+
+        # Unwrap rpy to [-pi, pi]		
+        rpy = numpy.add(xyzrpy[3:6], pi) % (2*pi) - pi
+        
         rpycheck = []
         for i in range(0, 3):
             if (Bw_rpy[i, 0] > Bw_rpy[i, 1] + EPSILON):
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -51,15 +51,20 @@ class TSR(object):
         pibound = (abs(Bw[3:6, :]) < numpy.pi + EPSILON)
         if pibound.any() is False:
             raise(ValueError('Rotations must be [-pi, pi]', pibound))
-        # We will now deliberately exceed the bound to maintain both:
-        # 1. Bw[i,1] > Bw[i,0] which is necessary for LBFGS-B
-        # 2. signed rotations, necessary for expressiveness
-        for rot_idx in range(3, 6):
-            if Bw[rot_idx, 0] > Bw[rot_idx, 1] + EPSILON:
-                Bw[rot_idx, 1] += 2*pi
+        
         self.T0_w = T0_w
         self.Tw_e = Tw_e
         self.Bw = Bw
+        
+        # We will now create a continuous version of the bound to maintain:
+        # 1. Bw[i,1] > Bw[i,0] which is necessary for LBFGS-B
+        # 2. signed rotations, necessary for expressiveness
+        Bw_cont = numpy.copy(Bw)
+        for rot_idx in range(3, 6):
+            if Bw_cont[rot_idx, 0] > Bw_cont[rot_idx, 1] + EPSILON:
+                Bw_cont[rot_idx, 1] += 2*pi
+        self._Bw_cont = Bw_cont
+        
         if manip is None:
             self.manipindex = -1
         elif type(manip) == openravepy.openravepy_int.Robot:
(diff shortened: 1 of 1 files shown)


src/prpy/tsr/tsr.py
```diff
@@ -51,15 +51,20 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
-            if Bw[rot_idx, 0] > Bw[rot_idx, 1] + EPSILON:
-                Bw[rot_idx, 1] += 2*pi
+        
         self.T0_w = T0_w
         self.Tw_e = Tw_e
         self.Bw = Bw
+        
+        # We will now create a continuous version of the bound to maintain:
+        # 1. Bw[i,1] > Bw[i,0] which is necessary for LBFGS-B
+        # 2. signed rotations, necessary for expressiveness
```
mkoval: I think I see what's going on here, but I'm confused by this comment. This code converts a signed rotation into a rotation that exceeds the [ -pi, pi ] range, so it does *not* "maintain signed rotations".

src/prpy/tsr/tsr.py
```diff
@@ -117,19 +122,17 @@ def is_valid(self, xyzrpy, ignoreNAN=False):
...
         Bw_rpy = self.Bw[3:6, :]
         rpycheck = []
         for i in range(0, 3):
-            if (Bw_rpy[i, 1] > pi + EPSILON):
+            if (Bw_rpy[i, 0] > Bw_rpy[i, 1] + EPSILON):
                 # An outer interval
-                rpycheck[i] = ((rpy[i] + EPSILON) >= Bw_rpy[i, 0]) or\
-                              ((rpy[i] + EPSILON) <= Bw_rpy[i, 1] - 2*pi)
+                rpycheck[i] = (((rpy[i] + EPSILON) >= Bw_rpy[i, 0]) or
+                               ((rpy[i] + EPSILON) <= Bw_rpy[i, 1]))
```
mkoval: I think this is incorrect with the "unwrap rpy to [-pi, pi]" check removed. This include *all values of `rpy`* that are not in the range `Bw_rpy[i, 1]` and `Bw_rpy[i, 0]`, even if they have wrapped around `2*pi` and should be considered in the excluded range.

src/prpy/tsr/tsr.py
```diff
@@ -186,9 +189,12 @@ def sample_xyzrpy(self, xyzrpy=NANBW):
...
             raise ValueError('xyzrpy must be within bounds', check)
 
-        return [self.Bw[i, 0] + (self.Bw[i, 1] - self.Bw[i, 0]) *
-                numpy.random.random_sample()
-                if numpy.isnan(x) else x for i, x in enumerate(xyzrpy)]
+        Bw_sample = [self._Bw_cont[i, 0] + (self._Bw_cont[i, 1] - self._Bw_cont[i, 0]) *
+                     numpy.random.random_sample()
+                     if numpy.isnan(x) else x for i, x in enumerate(xyzrpy)]
+        # Unwrap rpy to [-pi, pi]		
+        Bw_sample[3:6] = numpy.add(Bw_sample[3:6], pi) % (2*pi) - pi
```
mkoval: Out of curiosity, why is the explicit `numpy.add` call needed here?

src/prpy/tsr/tsr.py
psigen: This might be an outdated comment.  The "signed rotations" really mean being able to explicitly specify inner and outer angular bounds, which we are proposing to use reversed Bw ordering for.

The intuition is that the Bw[:,0] indicates a minimum angle, and then _increases_ from that to the Bw[:,1] value, regardless of angle wrapping.

src/prpy/tsr/tsr.py
psigen: The  "unwrap rpy to [-pi, pi]" hasn't been removed, it is on line 51 in the constructor.  `Bw` is in `[-pi, pi]`, `Bw_cont` is not.

src/prpy/tsr/tsr.py
psigen: No idea, copied from @siddhss5's code.

psigen:
@mkoval: Note that this is a PR into @siddhss5's `feature/tsr_distance`, not `master`, so some of the concerns you raised are related to the underlying branch.

This branch is just proposing to split out an internal `Bw_cont` which represents a continuous version of `Bw`, so that `Bw` can be enforced to lie within `[-pi. pi]`.

src/prpy/tsr/tsr.py
siddhss5: I believe @mkoval is referring to this particular line https://github.com/personalrobotics/prpy/pull/163#discussion-diff-36271837L121, where you've deleted the wrapping of `rpy` for some reason. I agree with him. That line is necessary, no?

siddhss5:
Overall, except for that one possible bug, I like this PR and will merge it. There's a related point @cdellin and I discussed about expressiveness of the input that should be a separate issue, but will affect this code.

src/prpy/tsr/tsr.py
siddhss5: I believe this was done by @rachelholladay. I'll track it down and figure it out.

rachelholladay: @siddh Using simply '+' threw a synatax error when I was testing. I believe I explicitly verified with you that numpy.add produced the intended functionality.

src/prpy/tsr/tsr.py
psigen: Oh, whoops, I see.  Yes, that wrap check should be in there, not sure why it got removed.

PR #166, opened 2015-08-20
Helper functions for computing body point acceleration twists

siddhss5:
I added some functionality in `prpy/util` to enable computation of acceleration twists of arbitrary body points, on their own, and also from an OpenRAVE trajectory.


src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
     """
     cspec = trajectory.GetConfigurationSpecification()
     return cspec.ExtractDeltaTime(trajectory.GetWaypoint(0)) is not None
+
+
+def jointposvelacc_from_traj(robot, traj, time):
```
psigen: Should probably keep with the OpenRAVE naming convention used elsewhere: `JointPosVelAccFromTraj`.

Is there a simpler name for this?  Maybe `JointMotionFromTraj`?

src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
...
+        trajdata = traj.Sample(t)
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 0))
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 1))
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 2))
+        pva_list.append(pva)
+
+    return pva_list
+
+
+def compute_acceleration_twist(link, world_pos, qd, qdd):
```
psigen: Similar to previous: `ComputeAccelerationTwist`.

src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
...
+    apos = numpy.dot(Jpos, qdd) + reduce(numpy.dot, [qd, Hpos, qd])
+
+    Jang = robot.CalculateAngularVelocityJacobian(link_index)
+    Hang = robot.ComputeHessianAxisAngle(link_index)
+    aang = numpy.dot(Jang, qdd) + reduce(numpy.dot, [qd, Hang, qd])
+
+    return numpy.hstack((apos, aang))
+
+
+def compute_acceleration_twist_from_traj(link, world_pos, traj, time):
```
psigen: Similar to previous: `ComputeAccelerationTwistFromTraj`.

src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
...
     return cspec.ExtractDeltaTime(trajectory.GetWaypoint(0)) is not None
+
+
+def jointposvelacc_from_traj(robot, traj, time):
+    """
+    Helper function to extract the joint position, velocity and acceleration
+    from an OpenRAVE trajectory.
+    @param robot The OpenRAVE robot
+    @param traj An OpenRAVE trajectory
+    @param time List of times in seconds
```
psigen: Maybe we can make this take both a `time` and a `times`, then throw in a:

```python
def jointposvelacc_from_traj(robot, traj, time=None, times=None):
    if not times and time:
        times = (time,)
    if times and time:
        throw ValueError('Cannot specify both time and times!')
```

src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
...
+
+    if any(time > duration):
+        raise ValueError('Input time {0:} exceeds duration {1:.2f}'
+                         .format(time, duration))
+
+    cspec = traj.GetConfigurationSpecification()
+    num_dofs = robot.GetDOF()
+    dof_indices = range(num_dofs)
+
+    pva_list = []
```
psigen: Should we do a preallocated `pva = numpy.zeros(len(time), 3)` instead?

src/prpy/util.py
```diff
@@ -749,3 +749,86 @@ def IsTimedTrajectory(trajectory):
...
+    for t in time:
+        pva = []
```
psigen: Should we do a preallocated `pva = numpy.zeros(len(dof_indices), 3)` instead?

src/prpy/util.py
siddhss5: I like `JointMotionFromTraj`. Would `JointStateFromTraj` be too confusing?

src/prpy/util.py
siddhss5: This is an excellent idea. There is still an issue with the return type which will be a `list` of `list` and have to be accessed as `foo[0][:]`, so it somewhat defeats the purpose, no?
It'd be so much nicer if this were just a `list` and accessible as `foo[:]`. But I fear that'll need a lot of code.

src/prpy/util.py
siddhss5: I didn't do this because some `pva` entries can be `None` and some weird behavior ensues:
```
In [101]: foo = numpy.zeros((6,3))

In [102]: foo[1] = None

In [103]: foo
Out[103]: 
array([[  0.,   0.,   0.],
       [ nan,  nan,  nan],
       [  0.,   0.,   0.],
       [  0.,   0.,   0.],
       [  0.,   0.,   0.],
       [  0.,   0.,   0.]])

In [104]: foo[1]
Out[104]: array([ nan,  nan,  nan])

In [105]: foo[1] is None
Out[105]: False
```

src/prpy/util.py
psigen: Hmm, I guess there's always the 10 cent solution:
```python
def JointMotionsFromTraj(robot, traj, times):
    [...]

def JointMotionFromTraj(robot, traj, time):
    return JointMotionsFromTraj(robot, traj, (time,))[0]
```

src/prpy/util.py
psigen: Hmm, good point.  Maybe @mkoval has some ideas on the best output format.

src/prpy/util.py
psigen: I would be OK with `JointState`, but I'm not too picky.

src/prpy/util.py
```diff
@@ -749,3 +749,168 @@ def IsTimedTrajectory(trajectory):
...
+
+def JointStatesFromTraj(robot, traj, times):
+    """
+    Helper function to extract the joint position, velocity and acceleration
+    from an OpenRAVE trajectory.
+    @param robot The OpenRAVE robot
+    @param traj An OpenRAVE trajectory
+    @param times List of times in seconds
+    @return pva_list List of list of position, velocity and acceleration
+                     at specified times. Inserts 'None' for unavailable fields
```
mkoval: This description doesn't make it clear that `pva_list` is a `list` of `list`s, where the `i`-th element is the `i`-th derivative of position.

src/prpy/util.py
```diff
@@ -749,3 +749,168 @@ def IsTimedTrajectory(trajectory):
...
+    num_dofs = robot.GetDOF()
+    dof_indices = range(num_dofs)
+
+    pva_list = []
+    for t in times:
+        pva = []
+        trajdata = traj.Sample(t)
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 0))
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 1))
+        pva.append(cspec.ExtractJointValues(trajdata, robot, dof_indices, 2))
```
mkoval: I'd prefer adding a `derivatives` parameter that includes a list of derivatives to compute. This would actually make the code shorter because these three lines could be replaced with a loop.

This would let you use this function to, e.g. only query positions or velocities. The default could be `[0, 1, 2]`, which would have the current

siddhss5: This is an excellent idea. But, I worry that `JointState` now might have variable size, depending on the derivatives that are requested. I fear this will lead to the same discomfort we have with _active_ joints. 

I feel like it's more natural if something not requested is `None` rather than be arbitrarily shortened.

src/prpy/util.py
siddhss5: Addressed in https://github.com/personalrobotics/prpy/commit/3092f503e8822cb6748ac9f52655ff3b171f9423. Still ambiguous?

src/prpy/util.py
siddhss5: Addressed in https://github.com/personalrobotics/prpy/commit/1fd2aeddc63d99cf86b996bcd0f7e1c46c402d64.

siddhss5:
@mkoval, take a look and tell me if you're happy with this.

src/prpy/util.py
```diff
@@ -749,3 +749,178 @@ def IsTimedTrajectory(trajectory):
...
+        raise ValueError('Input times {0:} exceed duration {1:.2f}'
+                         .format(times, duration))
+
+    cspec = traj.GetConfigurationSpecification()
+    num_dofs = robot.GetDOF()
+    dof_indices = range(num_dofs)
+
+    pva_list = []
+    for t in times:
+        pva = [None] * 3
```
mkoval: This should be:
```python
pva = [None] * len(derivatives)
```

src/prpy/util.py
```diff
@@ -749,3 +749,178 @@ def IsTimedTrajectory(trajectory):
...
+        trajdata = traj.Sample(t)
+        for d in derivatives:
+            pva[d] = cspec.ExtractJointValues(trajdata, robot,
+                                              dof_indices, d)
```
mkoval: This should be:
```python
for i, d in enumerate(derivatives):
    pva[i] = cspec.ExtractJointValues(trajdata, robot, dof_indices, d)
```

src/prpy/util.py
```diff
@@ -749,3 +749,178 @@ def IsTimedTrajectory(trajectory):
...
+def JointStateFromTraj(robot, traj, time):
+    """
+    Helper function to extract the joint position, velocity and acceleration
+    from an OpenRAVE trajectory.
+    @param robot The OpenRAVE robot
+    @param traj An OpenRAVE trajectory
+    @param time time in seconds
+    @return pva Position, velocity and acceleration
+                at specified time. Inserts 'None' for unavailable fields
+    """
```
mkoval: It would be nice to add `derivatives` as an argument here, too, and pass it down to `JointStatesFromTraj`.

src/prpy/util.py
```diff
@@ -749,3 +749,178 @@ def IsTimedTrajectory(trajectory):
...
+    @param robot The OpenRAVE robot
+    @param traj An OpenRAVE trajectory
+    @param times List of times in seconds
+    @param derivatives list of desired derivatives (only supports 0, 1, 2),
+                       defaults to [0, 1, 2]
+    @return pva_list List of list of position, velocity and acceleration
+                     at specified times.
+                     Inserts 'None' for unavailable or undesired fields
+                     The i-th element is the i-th derivative of position
+                     Of size |times| x 3
```
mkoval: This did fix my complaint, but it's incorrect with the new changes. The `i`-th element is the `derivative[i]`-th derivative of position and it's of size `|times| x |derivatives|`.

src/prpy/util.py
siddhss5: I don't think so. I thought we agreed that the `JointState` would always be of length `3` and just return `None` for the entries that are _unavailable_ or _undesired_, as the [docstring] (https://github.com/personalrobotics/prpy/pull/166#discussion-diff-37804723R765) suggests.
The new commit just enables being efficient about  _undesired_ derivatives.

Happy to discuss further if you're unhappy with it.

src/prpy/util.py
siddhss5: I find arbitrary derivative ordering a bit dangerous since there's no way to recover the derivative semantic from the `JointStates` list. I would actually consider doing something like:
```
for i in range(3):
    if i in derivatives:
        pva[i] = cspec.ExtractJointValues(trajdata, robot, dof_indices, i)
```

src/prpy/util.py
siddhss5: Alright, I propose a compromise: `JointState` to be ordered from `0 ... |derivatives|`.

psigen:
This API is getting weird... I think we're getting into design-by-committee problems here, guys.

Is there a way to split this function up such that @mkoval has a mechanism for polling arbitrary derivatives in any order and getting the returns in the same order while @siddhss5 has a mechanism for getting a contiguous range of derivatives in-order?

siddhss5:
If you were a crazy madman like @cdellin, you'd do something like this:
```
right_reduce = lambda fn,seq: reduce(lambda a,b: fn(b,a), reversed(seq))
result = right_reduce(np.dot, [qd,H,qd])
```
Or use [Haskell](https://wiki.haskell.org/Fold).

mkoval:
@psigen You can get the continuous range of derivatives by passing `range(n)`.

I'm open to taking `n` as the argument, instead of a list, and always returning the `0`-th through `n`-th derivatives. Does that eliminate the ambiguity that @siddhss5 is concerned about?

src/prpy/util.py
```diff
@@ -749,3 +749,197 @@ def IsTimedTrajectory(trajectory):
...
+
+    pva_list = []
+    for t in times:
+        pva = [None] * len(derivatives)
+        trajdata = traj.Sample(t)
+        for i in range(len(derivatives)):
+            if i in derivatives:
+                pva[i] = cspec.ExtractJointValues(trajdata, robot,
+                                                  dof_indices, i)
+        pva_list.append(pva)
```
mkoval: This looks like a bug: both loops are indexed by `i`, so the outer loop does nothing. If we're sticking with the `list` API, then this should be:
```python
for i, deriv in enumerate(derivatives):
    pva[i] = cspec.ExtractJointValues(trajdata, robot, dof_indices, deriv)
```

mkoval:
The changes look good besides the one comment I posted about `JointStatesFromTraj`.

@siddhss5 @psigen Are you both happy with this API? Or do you want to go to the single integer, instead?

cdellin:
@mkoval, you could accept either integers or sequences, and interally treat `n` as `range(n)` ...

mkoval:
@cdellin: I made that suggestion in response to @siddhss5's concern:

> I find arbitrary derivative ordering a bit dangerous since there's no way to recover the derivative semantic from the JointStates list.

Requiring the user to specify `n` makes the `JointState` list unambiguous because you **always** know that the `i`-th element corresponds to the `i`-th derivative. That is not true if we allow **any** way to specify a `list` that is not of the form `range(n)`.

cdellin:
@mkoval, agreed.  I'm personally happy with any API, although I think that having some way to only calculate a higher derivative might be useful for efficiency.

If the requirement that the `i`-th element of the returned `JointState` sequence corresponds to the `i`-th derivative is important, but we still want a way to only calculate a subset of the derivatives for efficiency, then the returned sequence could be padded with `None`s.  Alternatively, the `JointState` could be a `dict` keyed on the derivative index, or a custom class.

psigen:
I suggest we just split it into two calls, one with arbitrary lists, and a
trivial helper that takes an int and just calls the other with range(n).

siddhss5:
> This discussion goes beyond my PR and I suggest you start another branch to worry about  such intricacies. I'll address any perceived bugs right now and close this PR.

In general I'm happy with API discussion but remember to understand the scope of a PR before meandering.

mkoval:
I think that [my comment above](https://github.com/personalrobotics/prpy/pull/166#discussion_r38239139) is a bug. Can you check that before we merge this?

psigen:
Agreed, let's just address @mkoval's [bug](https://github.com/personalrobotics/prpy/pull/166#discussion_r38239139) and we'll merge then move the rest of this into a post-PR issue to discuss.

mkoval:
I'm still happy to merge this once the bug is fixed.

@psigen or @siddhss5 can you take a shot at fixing it?

mkoval:
There are conflicts that must be resolved before I can merge this pull request:

> This branch has conflicts that must be resolved.

psigen:
@mkoval :grumble grumble: fixed merge.

PR #171, opened 2015-09-10
Fixed MoveUntilTouch on HERB

mkoval:
This pull request adds support for setting OWD execution options (via `SetTrajectoryExecutionOptions`) on trajectories via the recent changes in https://github.com/personalrobotics/or_owd_controller/pull/4. It also updates `MoveUntilTouch` to use this functionality.

It is not currently possible to detect whether the trajectory was aborted or not. This needs to be fixed before I can merge these changes.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -340,7 +340,13 @@ class Robot(openravepy.Robot):
                             cloned_robot, shortcut_path, defer=False,
                             **smoothing_options)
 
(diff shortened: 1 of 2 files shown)


mkoval:
I added the functionality necessary to check for an abort to HerbPy in https://github.com/personalrobotics/herbpy/pull/41. We should be back to the same level of functionality we had before the trajectory execution refactor.

PR #173, opened 2015-09-11
simple fix for simulation of move until touch

sjavdani:


mkoval:
:+1: Thanks @sjavdani!

PR #175, opened 2015-09-13
Deprecated the GetVelocityLimits override on WAM

### LINKED ISSUE #172: GetVelocityLimits fails
A call to ```robot.right_arm.GetVelocityLimits()``` throws the following error:
```
/home/herb_home/shared/block_ws/src/prpy/src/prpy/base/wam.pyc in GetVelocityLimits(self, openrave, owd)
    181             args  = [ 'GetSpeed' ]
    182             args_str = ' '.join(args)
--> 183             owd_speed_limits_all = self.controller.SendCommand(args_str)
    184             #if we get nothing back, e.g. if the arm isn't running, return openrave lims
    185             if owd_speed_limits_all is None:

openrave_exception: openrave (CommandNotSupported): failed to find command 'GetSpeed' in interface OWDController prpy /right/owd
```

mkoval:
This pull request fixes #172 when running on the real robot. I removed the option to query OWD's velocity limits, since the underlying `GetSpeed` call is not implemented in `or_owd_controller`.

@jeking04 Do we use this anywhere in `table_clearing` or `block_sorting`?

### DIFF
--- a/src/prpy/base/wam.py
+++ b/src/prpy/base/wam.py
@@ -30,6 +30,7 @@
 
 import numpy
 import openravepy
+import warnings
 from manipulator import Manipulator
 from prpy.clone import Clone
 from .. import util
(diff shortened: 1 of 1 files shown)


jeking04:
Nope I don't think so.

mkoval:
Ok, I'm merging this then.

Issue #158, opened 2015-07-14
Remove default implementation of has_planning_method() from MetaPlanner


cdellin:
Mike wanted me to make an issue to remind him to do this. Below text is from him (on slack)

---

the default implementation of has_planning_method seems pretty bogus:

    def has_planning_method(self, method_name):
        for planner in self._planners:
            if planner.has_planning_method(method_name):
                return True

        return False

probably left over for backwards compatability. can you file an issue so I remember to remove it?

MetaPlanner inherits from Planner (which it should), but Planner has some default implementations that may not work on meta-planners

this all got messed up when we refactored the meta-planners...

mkoval:
Fixed in #176.

PR #176, opened 2015-09-13
Removed default methods on MetaPlanner.

mkoval:
The `get_planning_method_names()` and `has_planning_method()` vary between MetaPlanners and, thus, should be implemented in derived classes. @cdellin caught this in #158.


mkoval:
This does not break `Sequence`, which is the only `MetaPlanner` we use by default now, so I am merging this.

PR #184, opened 2015-09-16
Improvements to VectorFieldPlanner

mkoval:
This pull request makes several major improvements to `VectorFieldPlanner` to address #99:

- Decouple the integration step size from the collision and constraint checking resolution.
- Use a numerical integrator (from `scipy.integrate`) instead of fixed-size Euler integration.
- Ignore velocity constraints when defining the vector field.
- All checks are still done during the integration; not as a post-processing step.

It also adds some generally-useful functionality:

- `CACHE_AND_TERMINATE` flag for `TERMINATE`ing while still `CACHE`ing the current configuration
- `ComputeUnitTiming` computes the arclength parameterization of a path or trajectory.
- `GetCollisionCheckPts` generates a sequence of `(t, q)` pairs that are within DOF resolution using the same bisection method as `GreedyIK`
- `ComputeJointVelocityFromTwist` can optionally ignore the robot's velocity limits; e.g. to compute an answer up to scale

**There is one blocking issue:** I can't figure out how to handle numerical precision in `GetCollisionCheckPts` when starting partway through a trajectory. I don't see why this is any different than starting at zero, but it causes the bisection test to get stuck in an infinite loop. I'd appreciate another set of eyes to figure out what I"m doing wrong.

This includes some major changes and has some nasty edge cases. I'd really like @jeking04, @psigen, and @siddhss5 to look at this before I merge it.

### DIFF
--- a/src/prpy/planning/exceptions.py
+++ b/src/prpy/planning/exceptions.py
@@ -37,6 +37,35 @@ class CollisionPlanningError(PlanningError):
             return '<unknown>'
 
 
+class JointLimitError(PlanningError):
+    def __init__(self, robot, dof_index, dof_value, dof_limit, description):
+        self.robot = robot
+        self.dof_index = dof_index
+        self.dof_value = dof_value
+        self.dof_limit = dof_limit
+
+        joint = robot.GetJointFromDOFIndex(dof_index)
+        if dof_value < dof_limit:
+            direction = 'lower'
+            comparison = '<'
+        else:
+            direction = 'upper'
+            comparison = '>'
+
+        super(JointLimitError, self).__init__(
+            'Robot "{robot_name:s}" joint "{joint_name:s} axis {joint_axis:d}'
+            ' violates {direction:s} {description:s} limit:'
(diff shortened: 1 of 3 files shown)


mkoval:
Also tagging @aaronjoh because I know nothing about numerical integration.

siddhss5:
Oh yikes! My daughter merged this by accident on the phone! Dammit.

src/prpy/planning/exceptions.py
```diff
@@ -37,6 +37,35 @@ def _get_link_str(link):
...
+            direction = 'lower'
+            comparison = '<'
+        else:
+            direction = 'upper'
+            comparison = '>'
+
+        super(JointLimitError, self).__init__(
+            'Robot "{robot_name:s}" joint "{joint_name:s} axis {joint_axis:d}'
+            ' violates {direction:s} {description:s} limit:'
+            ' {dof_value:.5f} {comparison:s} {dof_limit:.5f}'.format(
```
psigen: Need to indicate which value is which in the comparison.

PR #188, opened 2015-09-22
Adding logic to make IsAtTrajectoryStart work for affine dofs.

jeking04:
Here is an initial attempt at this. I imagine it will need some iteration.


src/prpy/util.py
```diff
@@ -710,29 +716,56 @@ def IsAtTrajectoryStart(robot, trajectory):
...
-        axis = dof_index - joint.GetDOFIndex()
-
-        # If any joint deviates too much, return False.
-        delta_value = abs(joint.SubtractValue(traj_value, robot_value, axis))
-        if delta_value > dof_resolution:
+    needs_base = HasAffineDOFs(cspec)
+    needs_joints = HasJointDOFs(cspec)
+    
+    if needs_base and needs_joints:
+        raise ValueError('Trajectories with affine and joint DOFs are not supported')
```
mkoval: We also need to check if the trajectory is empty. Otherwise, some of the code below will crash.

src/prpy/util.py
```diff
@@ -710,29 +716,56 @@ def IsAtTrajectoryStart(robot, trajectory):
...
+
+        # Compare translation distance
+        trans_delta_value = abs(current_pose[:2] - traj_start[:2])
+        trans_resolution = robot.GetAffineTranslationResolution()[:2]
+        if trans_delta_value[0] > trans_resolution[0] or \
+           trans_delta_value[1] > trans_resolution[1]:
+            return False
+        
+        # Compare rotation distance
+        rot_delta_value = (current_pose[2] - traj_start[2] + numpy.pi) % (2.*numpy.pi) - numpy.pi
```
mkoval: I think this comparison object may return a negative `rot_delta_value` if `current_pose[2]` and/or `traj_start[2]` are very large. This is also a very error-prone piece of code to write. Can we move this to a helper function and use an implementation from a trusted library?

src/prpy/util.py
mkoval: This will also fail if `traj.GetEnv() != robot.GetEnv()`. We should check for this here and throw an exception.

mkoval:
This pull request actually looks pretty good! I expected that it would be quite a bit hairier after our previous experiences with affine DOFs. :sweat: 

I added notes about two small edge cases above.

src/prpy/util.py
jeking04: Done.

src/prpy/util.py
jeking04: Sounds good. I am having trouble finding a trusted library to use. any suggestions?

psigen: Also, someone needs to make a decision about what this angle wrapping should be, because we are already having issues where some of the code wraps to `[pi, pi]` while other parts wrap to `[-pi,pi)`, while I think `(-pi,pi]` is probably the most intuitive non-redundant interval.

See: https://github.com/personalrobotics/prpy/issues/190

mkoval: @jeking04 Can you modify this to use the the `wrap_to_interval` function that @psigen added in #190 to do this check?

I believe we can compute distance as `abs(wrap_to_interval(theta2 - theta1))`. Does that make sense?

mkoval:
The only pending change is to modify the orientation comparison code to use `wrap_to_interval`.

@jeking04 If you make that change, I can go ahead and merge this.

jeking04:
Okay @mkoval, have a look. I made the change and moved wrap_to_interval into prpy util.

mkoval:
LGTM. Thanks!

PR #189, opened 2015-09-22
Block sorting

jeking04:
This pulls in the perception pipeline and a few other enhancements made during the block sorting demo.


src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+            box_max_pt.x = box_max[0];
+            box_max_pt.y = box_max[1];
+            box_max_pt.z = box_max[2];
+
+            service = rospy.ServiceProxy(service_name, FindBlocks);
+            response = service(cloud_topic, segment_planes, num_planes, plane_distance, segment_depth, min_depth, max_depth, cluster_tolerance, min_cluster_size, max_cluster_size, segment_box, box_min_pt, box_max_pt);
+            return response.blocks;
+        except rospy.ServiceException, e:
+            print "Service call failed: %s"%e
+            return []
```
mkoval: We should replace these `print` calls with `logging` messages.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+        #tray_aabb = tray.ComputeAABB()
+        #env.Add(tray)
+
+        #xpose = table
+        #tray_in_table = numpy.array([[1., 0., 0., -0.7607], # right edge - -.99575 + .235
+        #                             [0., 0., 1., table_height + 0.01],
+        #                             [0.,-1., 0., 0.],
+        #                             [0., 0., 0., 1.]])
+        #tray_in_world = numpy.dot(table.GetTransform(), tray_in_table)
+        #tray.SetTransform(tray_in_world)
```
mkoval: I don't think this code is necessary now that we perceive the table. Can we delete it?

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+        """
+        if len(blocks) == 0:
+            # Add all blocks
+        
+            env = robot.GetEnv()
+
+            # Get the pr-ordata package path
+            import prpy.util
+            data_dir = prpy.util.FindCatkinResource('pr_ordata', 'data')
+            block_path = os.path.join(data_dir, 'objects', 'block.kinbody.xml')
```
mkoval: `openrave_catkin` automatically adds `pr_ordata/data` to the `OPENRAVE_DATA` path. We should load the block as `objects/block.kinbody.xml` and **should not** resolve an absolute path using `FindCatkinResource`.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+                valid = False
+                while not valid:
+                    rand_name = 'block' + `int(numpy.random.randint(1,10000))`
+                    valid = env.GetKinBody(rand_name) is None
+                
+                block.SetName(rand_name)
+                env.Add(block)
+                blocks.append(block)
+
+        return blocks
```
mkoval: The environment should be locked around all of this.

src/prpy/planning/retimer.py
```diff
@@ -154,6 +154,15 @@ def __init__(self, do_blend=True, do_shortcut=True, blend_radius=0.5,
...
 
+    @PlanningMethod
+    def RetimeTrajectory(self, robot, path, options=None, **kw_args):
+        from copy import deepcopy
+        if options is None:
+            options = {}
+        new_options = deepcopy(options)
+        new_options['time_limit'] = kw_args.get('timelimit', 3.)
+        return super(HauserParabolicSmoother, self).RetimeTrajectory(
+            robot, path, options=new_options, **kw_args)
```
mkoval: @jeking04 Do you remember why this is necessary?

src/prpy/planning/vectorfield.py
```diff
@@ -357,7 +357,7 @@ def fn_callback(t, q):
         output_path.Insert(0, path.GetWaypoints(0, cached_index), cspec)
 
         # Add a segment for the feasible part of the last integration step.
-        output_path.Insert(output_path.GetNumWaypoints() - 1,
+        output_path.Insert(output_path.GetNumWaypoints(),
```
mkoval: Is this correct? Or is there still an off-by-one-error?

I always get confused about the `Insert` method on `Trajectory`s.

mkoval:
Overall, this looks pretty good. I flagged a couple minor issues. @jeking04 could you take a look at them before I merge this?

src/prpy/base/wam.py
```diff
@@ -31,6 +31,7 @@
 import logging
 import numpy
 import openravepy
+import time
```
mkoval: This `import` appears to be unused.

src/prpy/base/wam.py
```diff
@@ -292,17 +293,22 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
...
         # tell the controller to stop on force/torque input.
         if not manipulator.simulated:
-            manipulator.SetTrajectoryExecutionOptions(traj, stop_on_ft=True,
+            manipulator.SetTrajectoryExecutionOptions(path, stop_on_ft=True,
                 force_direction=force_direction, force_magnitude=max_force,
                 torque=max_torque)
 
             manipulator.hand.TareForceTorqueSensor()
 
+            vl = manipulator.GetVelocityLimits()
```
mkoval: We should replace this manual saving logic with a `KinBodyStateSaver` with the `Save_JointMaxVelocityAndAcceleration` option. We should also make this scale factor (currently 0.25)` a parameter.

src/prpy/perception/apriltags.py
```diff
@@ -0,0 +1,123 @@
...
+import logging
+from base import PerceptionModule, PerceptionMethod
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+class ApriltagsModule(PerceptionModule):
+    
+    def __init__(self, marker_topic, marker_data_path, kinbody_path,
+                 detection_frame, destination_frame):
```
mkoval: Please add a `__docstring__` to this function, including the arguments.

src/prpy/perception/apriltags.py
```diff
@@ -0,0 +1,123 @@
...
+        self.marker_data_path = marker_data_path
+        self.kinbody_path = kinbody_path
+        self.detection_frame = detection_frame
+        self.destination_frame = destination_frame
+        
+        
+    def __str__(self):
+        return 'Apriltags'
+
+    def DetectObjects(self, robot, object_names, **kw_args):
```
mkoval: None of these functions seem to use `robot`. Instead, they always query `robot.GetEnv()`. Can we change all of these functions to take an `Environment` directly?

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+import tabletop_perception_tools.msg
+from tabletop_perception_tools.srv import *
+import rospy
+import tf
+import geometry_msgs.msg
+from tf.transformations import quaternion_matrix,euler_from_matrix,euler_matrix
+
+from base import PerceptionModule, PerceptionMethod
+
+table_z_offset = -0.02
```
mkoval: We should make this an optional parameter in the function.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,201 @@
...
+                table_aabb = ComputeEnabledAABB(table)
+            z = table_aabb.pos()[2] + table_aabb.extents()[2] + table_z_offset #OFFSET SET AT TOP
+
+            for b in detected_blocks:
+
+                block = env.ReadKinBodyXMLFile(block_path)
+
+                for link in block.GetLinks():
+                    for geometry in link.GetGeometries():
+                        geometry.SetDiffuseColor(numpy.array([b.avg_color.r,b.avg_color.g,b.avg_color.b,b.avg_color.a]))
```
mkoval: Something is off with the spacing here.

src/prpy/perception/rock_module.py
```diff
@@ -0,0 +1,140 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+import logging
+from base import PerceptionModule, PerceptionMethod
+
+logger = logging.getLogger(__name__)
+
+class RockModule(PerceptionModule):
+
+    def __init__(self, kinbody_path, detection_frame, destination_frame):
```
mkoval: Missing `__docstring__`.

src/prpy/perception/rock_module.py
```diff
@@ -0,0 +1,140 @@
...
+        poses = rock_detector.SearchTabletop(obj_name)
+    
+        # Use the poses to generate new kinbodies
+        kinbodies = rock_detector.KinBodiesFromPoses(obj_names, poses)
+
+        return kinbodies
+
+    @PerceptionMethod
+    def DetectObjectNearPose(self, robot, obj_name, pose, 
+                             search_extents=[-0.1, -0.1, 0.],
```
mkoval: We should [*never* set default values to mutable objects](https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/). This should be changed to `None` and an `if is None` check in the function body.

src/prpy/perception/simulated.py
```diff
@@ -0,0 +1,52 @@
...
+import logging
+from base import PerceptionModule, PerceptionMethod
+
+logger = logging.getLogger(__name__)
+
+class SimulatedPerceptionModule(PerceptionModule):
+    
+    def __init__(self):
+        
+        super(SimulatedPerceptionModule, self).__init__()
```
mkoval: Please fix spacing.

src/prpy/base/wam.py
jeking04: True. We added the import as part of a change that was eventually removed. However, when I removed that change I noticed the ServoTo function uses the time module without importing it. So I kept the import. I have now moved it down into the ServoTo function where it is needed.

src/prpy/base/wam.py
jeking04: Done and Done.

src/prpy/perception/block_detector.py
jeking04: Done.

src/prpy/perception/rock_module.py
jeking04: Done.

src/prpy/perception/simulated.py
jeking04: I made a change. Let me know if this is not the fix your were referring to.

src/prpy/planning/retimer.py
jeking04: Hauser smoother expects a parameter named 'time_limit'.  Our prpy planning convention is to use 'timelimit'.

src/prpy/perception/apriltags.py
Shushman: Addressed

src/prpy/perception/apriltags.py
Shushman: Addressed.

src/prpy/perception/rock_module.py
Shushman: Addressed

src/prpy/perception/block_detector.py
Shushman: Addressed (I think?)

src/prpy/base/wam.py
```diff
@@ -236,6 +238,7 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
         @param max_force maximum force in Newtons
         @param max_torque maximum torque in Newton-Meters
         @param ignore_collisions collisions with these objects are ignored when planning the path, e.g. the object you think you will touch
+        @param velocity_limit_scale A multiplier to use to scale velocity limits when executing MoveUntilTouch ( < 1 in most cases).           
         @param **kw_args planner parameters
```
mkoval: Please wrap these lines to 80 characters to be PEP-8 compliant.

src/prpy/perception/apriltags.py
```diff
@@ -0,0 +1,139 @@
...
+    def __init__(self, marker_topic, marker_data_path, kinbody_path,
+                 detection_frame, destination_frame):
+        """
+        This initializes an April Tags detector.
+        Args:
+            marker_topic (string): The ROS topic to read markers from. Typically the output topic for April Tags
+            marker_data_path(string): The json file where the association between tag and object is stored
+            kinbody_path: The path to the folder where kinbodies are stored
+            detection_frame: The TF frame of the camera
+            destination_frame: The desired world TF frame
```
mkoval: Please follow the same syntax we use elsewhere in `prpy` for argument names, argument types, and return value types. See [`PostProcessPath`](https://github.com/personalrobotics/prpy/blob/master/src/prpy/base/robot.py#L188) for a good example.

src/prpy/perception/apriltags.py
```diff
@@ -0,0 +1,139 @@
...
+        Use the apriltags service to detect objects and add them to the
+        environment
+        """
+        try:
+            # Allow caller to override any of the initial parameters
+            # loaded into the module
+            if 'marker_data_path' in kw_args:
+                marker_data_path = kw_args['marker_data_path']
+            else:
+                marker_data_path = self.marker_data_path
```
mkoval: I would prefer to make all of these values in `kwargs` explicit keyword arguments in the function. All of these checks could then be replaced with something like this:
```python
if kinbody_path is None:
    kinbody_path = self.kinbody_path`.
```

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,166 @@
...
+from base import PerceptionModule, PerceptionMethod
+
+import logging
+logger = logging.getLogger(__name__)
+
+table_z_offset = -0.02
+
+class BlockDetector(PerceptionModule):
+
+    def __init__(self,point_cloud_topic,detection_frame,destination_frame):
```
mkoval: Missing `__docstring__`.

src/prpy/perception/block_detector.py
mkoval: Actually, it's just Github being silly. Disregard this.

src/prpy/perception/block_detector.py
Shushman: Addressed - no data_dir used now.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,166 @@
...
+                        b.pose.orientation.w]))
+                block_pose[0,3] = b.pose.position.x
+                block_pose[1,3] = b.pose.position.y
+                block_pose[2,3] = b.pose.position.z
+
+                self.listener.waitForTransform(
+                        self.detection_frame,
+                        self.destination_frame,
+                        rospy.Time(),
+                        rospy.Duration(5))
```
mkoval: Can you make this timeout a parameter with a default value?

src/prpy/perception/block_detector.py
mkoval: Please disregard. The user should lock the environment around this call.

src/prpy/perception/rock_module.py
```diff
@@ -0,0 +1,150 @@
...
+            detection_frame(string): The TF frame of the camera
+            destination_frame(string): The TF frame of the world
+        """
+        super(RockModule, self).__init__()
+        self.kinbody_path = kinbody_path
+        self.detection_frame = detection_frame
+        self.destination_frame = destination_frame
+
+    def __str__(self):
+        return 'RockModule'
```
mkoval: I'd prefer `return self.__class__.__name__` here, but this is just a nitpick.

src/prpy/perception/rock_module.py
```diff
@@ -0,0 +1,150 @@
...
+class RockModule(PerceptionModule):
+
+    def __init__(self, kinbody_path, detection_frame, destination_frame):
+        """
+        Initialize the rock module
+        Args:
+            kinbody_path(string): The path to the directory that has kinbodies of objects
+            detection_frame(string): The TF frame of the camera
+            destination_frame(string): The TF frame of the world
+        """
```
mkoval: `__docstring__` format (see above).

src/prpy/planning/retimer.py
mkoval: Can we just fix `HauserParabolicSmoother` to use `timelimit`, instead?

src/prpy/base/wam.py
```diff
@@ -292,16 +295,19 @@ def MoveUntilTouch(manipulator, direction, distance, max_distance=None,
...
+            manipulator.SetTrajectoryExecutionOptions(path, stop_on_ft=True,
                 force_direction=force_direction, force_magnitude=max_force,
                 torque=max_torque)
 
             manipulator.hand.TareForceTorqueSensor()
 
             try:
-                robot.ExecutePath(path)
+                with robot.CreateRobotStateSaver(Robot.SaveParameters.JointMaxVelocityAndAcceleration):
+                    manipulator.SetVelocityLimits(velocity_limit_scale*vl, 0.5)
```
mkoval: It looks like the `vl` variable is now undefined.

src/prpy/perception/apriltags.py
Shushman: But the method that overrides `@PerceptionMethod`, the `DetectObjects` method has `**kwargs` as a param. It can't call `_DetectObjects` inside it with `**kwargs` if this change is made, right?

mkoval: It should be fine. You may, however, still need to include `**kwargs` at the end of the list of explicitly specified keyword arguments to catch any unexpected values.

src/prpy/planning/retimer.py
jeking04: Sorry, I wasn't clear on my previous comment. Hauser smoother expects the option to be named ```time_limit``` internally. This line allows us to have the ```HauserParabolicSmoother``` use ```timelimit``` and transfer it to the ```time_limit``` option to be passed on to the actual Hauser smoother/planner.

src/prpy/planning/vectorfield.py
jeking04: Looking at the openrave implementation of ```GenericTrajectory```, ```Insert``` puts the element at the specified index:

https://github.com/rdiankov/openrave/blob/b8fa1bb1fad5a86d373cc5d286e925f694540ea4/src/libopenrave-core/generictrajectory.cpp#L103

Here we want to append to the end of the trajectory, so passing ```output_path.GetNumWaypoints()``` should be correct.

src/prpy/base/wam.py
jeking04: Addressed.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,172 @@
...
+        except rospy.ServiceException, e:
+            logging.error("Service call failed: %s", str(e))
+            return []
+
+
+    @PerceptionMethod
+    def DetectBlocks(self, robot, table, blocks=None,timeout=10, **kw_args):
+        """
+        Place blocks on the table
+        """
```
mkoval: Please add documentation for the arguments and the return value (see above).

src/prpy/planning/retimer.py
mkoval: Okay, that makes sense.

src/prpy/planning/vectorfield.py
mkoval: That makes sense.

mkoval:
There are three minor issues remaining before we can merge this:

1. There are merge conflicts that must be resolved (via `git merge master` locally)
2. [Cleanup one function](https://github.com/personalrobotics/prpy/pull/189/files#r41671646)
3. [Add one missing `__docstring__`](https://github.com/personalrobotics/prpy/pull/189/files#r41699231)

@Shushman Can you address these?

src/prpy/planning/retimer.py
psigen: Um.. we did also write `or_parabolicsmoother`.  Perhaps we should just change it to be `timelimit` there?

https://github.com/personalrobotics/or_parabolicsmoother/blob/master/include/or_parabolicsmoother/ParabolicSmootherParameters.h#L117

mkoval: I agree that we should make it consistent (one way or the other). However, we've been living with the name mismatch for months and the change has nothing to do with this pull request. Let's keep this pull request as is and change the name in a separate PR.

jeking04:
Okay, @mkoval, I think we addressed all 3 above. After merging master into this branch I ran the unittests. The planning tests passed but the TSR test failed. Switching back to master I see it fails there too. So I think we should fix it under a different PR.

src/prpy/perception/block_detector.py
```diff
@@ -0,0 +1,180 @@
...
+                block_in_world_corrected = euler_matrix(ax,ay,az)
+                block_in_world[0:3,0:3] = block_in_world_corrected[0:3,0:3]
+                
+
+                #Set block name - should we change?
+                block.SetTransform(block_in_world)
+                valid = False
+                while not valid:
+                    rand_name = 'block' + `int(numpy.random.randint(1,10000))`
+                    valid = env.GetKinBody(rand_name) is None
```
mkoval: `env.Add` takes an optional second argument `anonymous` that will append an integer to the end of the `KinBody`'s name to make it unique. Please replace this loop with that functionality.

Shushman: Addressed

PR #191, opened 2015-09-24
Adds back missing CONSTRAINED trajectory tag.

psigen:
For some reason, the CONSTRAINED trajectory tag was lost during the vectorfield refactor.  This PR adds it back.

### DIFF
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -361,7 +361,11 @@ class VectorFieldPlanner(BasePlanner):
             path.Sample(t_cache), cspec)
 
         # Flag this trajectory as constrained.
-        # TODO: Flag this trajectory as timed.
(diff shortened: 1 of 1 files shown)


mkoval:
We've determined that using the time information from the integration is a bad idea. It's generated using s first order approximation, so the result isn't pretty when you try to execute it with real dynamics. Can you remove that comment?

Also, shouldn't we flag the output as smooth? The waypoints are quite sense, so it should be safe to simply interpolate between them.

psigen:
I don't know? I don't know what the difference between `SMOOTH` and `CONSTRAIN` is intended to be, 

I thought `SMOOTH` required valid timing but there's no comments on any of them so I just put whatever was there before.

mkoval:
`SMOOTH` was intended to mean that the waypoints are close enough together that you can approximate derivatives at the waypoints using divided differences. I.e. we can safely fit a spline without collision checking.

PR #193, opened 2015-09-24
Added documentation for `prpy.base.Tags`.

psigen:
Added some docstrings for the various tags to explain their meanings.

@mkoval: Did I get these right?


mkoval:
That looks good. Is that really the right position for variable `docstring`s?

psigen:
Sort of: http://stackoverflow.com/q/12698102

It's either that or a massive block added to the class docstring, which is arguably less useful here.

Issue #190, opened 2015-09-23
TSR Bw bounds for rotations


jeking04:
I would like to setup a TSR that is unbounded in yaw (or roll or pitch). Previously, I would do this by setting the Bw matrix up as follows:

```
Bw[5,:] = [-numpy.pi, numpy.pi]
```
With the new implementation of TSRs, this gets adjusted so that ```Bw[5,:] = [-numpy.pi, -numpy.pi]```, resulting in no variation in the yaw component when sampling from the TSR.

I have been fixing this by setting the Bw matrix up as follows:
```
Bw[5,:] = [-numpy.pi, numpy.pi - 0.0001]
```
Is there a different way I should be specifying these components?

psigen:
This is an ongoing bug in TSR bounds:

@siddhss5 and I are trying to figure out how to handle TSR boundaries, which in various locations are now wrapped or unwrapped to `[-pi, pi)` and `[-pi, pi]`

The short term hack is, as you mentioned, to subtract epsilon.

psigen:
@jeking04: I think I might have a fix for this here, can you take a look when you get a chance?
https://github.com/personalrobotics/prpy/pull/194

mkoval:
This was fixed in #194.

PR #194, opened 2015-09-29
Patched TSR wrapping to use Bw_cont and added TSR unit test.

psigen:
This PR addresses https://github.com/personalrobotics/prpy/issues/190 by changing the wrapping function to be bounds-aware.  This allows wrapping into the `Bw` interval for checks, and wrapping into the `[-pi, pi)` interval for samples, and also fixes various bugs with zero-measure intervals wrapping incorrectly.

The PR also adds a unit test for TSRs, although it is very simple right now.

### DIFF
--- a/src/prpy/tsr/tsr.py
+++ b/src/prpy/tsr/tsr.py
@@ -36,6 +36,26 @@ NANBW = numpy.ones(6)*float('nan')
 EPSILON = 0.001
 
 
+def wrap_to_interval(angles, lower=-pi):
+    """
+    Wraps an angle into a semi-closed interval of width 2*pi.
+
+    By default, this interval is `[-pi, pi)`.  However, the lower bound of the
+    interval can be specified to wrap to the interval `[lower, lower + 2*pi)`.
+
+    If `lower` is an array the same length as angles, the bounds will be
+    applied element-wise to each angle in `angles`.
+
+    See: http://stackoverflow.com/a/32266181
+
+    @param angles an angle or 1D array of angles to wrap
+    @type  angles float or numpy.array
+    @param lower optional lower bound on wrapping interval
+    @type  lower float or numpy.array
+    """
+    return (angles - lower) % (2*pi) + lower
+
+
 class TSR(object):
     """ A Task-Space-Region (TSR) represents a motion constraint. """
     def __init__(self, T0_w=None, Tw_e=None, Bw=None,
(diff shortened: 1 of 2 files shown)


src/prpy/tsr/tsr.py
```diff
@@ -46,21 +66,25 @@ def __init__(self, T0_w=None, Tw_e=None, Bw=None,
...
-        for rot_idx in range(3, 6):
-            if Bw_cont[rot_idx, 0] > Bw_cont[rot_idx, 1] + EPSILON:
-                Bw_cont[rot_idx, 1] += 2*pi
+        Bw_cont = numpy.copy(self.Bw)
+
+        Bw_interval = Bw_cont[3:6, 1] - Bw_cont[3:6, 0]
+        Bw_interval = numpy.minimum(Bw_interval, 2*pi)
+
+        Bw_cont[3:6, 0] = wrap_to_interval(Bw_cont[3:6, 0])
+        Bw_cont[3:6, 1] = Bw_cont[3:6, 0] + Bw_interval
```
mkoval: This can produce values in the range `[ -pi, 3*pi ]`. Is that intentional?

src/prpy/tsr/tsr.py
```diff
@@ -175,8 +199,8 @@ def rpy_within_bounds(rpy, Bw):
         @param Bw bounds on rpy
         @return check a (3,) vector of True if within and False if outside
         """
-        # Unwrap rpy to [-pi, pi].
-        rpy = (numpy.array(rpy) + pi) % (2*pi) - pi
+        # Unwrap rpy to Bw_cont.
+        rpy = wrap_to_interval(rpy, lower=Bw[:, 0])
```
mkoval: I thought the purpose of creating `Bw_cont` was to unify the treatment of all angular ranges. However, in this code, we ignore `Bw_cont` and implement separate logic for inner and outer intervals. Is this intentional?

tests/tsr/test_tsr.py
```diff
@@ -0,0 +1,36 @@
...
+              [0.,           0.],  # Z
+              [pi,        3.*pi],  # roll
+              [pi/2.,   3*pi/2.],  # pitch
+              [-3*pi/2., -pi/2.]]  # yaw
+        tsr = TSR(Bw=Bw)
+        s = tsr.sample_xyzrpy()
+
+        Bw = numpy.array(Bw)
+        self.assertTrue(numpy.all(s >= Bw[:, 0]))
+        self.assertTrue(numpy.all(s <= Bw[:, 1]))
```
mkoval: It would be nice to have a similar test, possibly using the same TSR, on `contains`. I took a look through the implementation and I'm not really sure whether the various Euler angle conversion functions correctly handle the new ranges of bounds (especially since we can produce values from `[ 0, 3*pi ]`).

src/prpy/tsr/tsr.py
psigen: How does this happen?

src/prpy/tsr/tsr.py
psigen: At the end of the day, the user only knows about the `Bw` range, which they have specified and may be at some crazy value or loop multiple times or whatever.  So it seems like we should check that we are in this range.

If we can guarantee through properties or something else that  `angle \in Bw_cont iff Bw` is always true, then we can check `Bw_cont` instead.

mkoval:
This pull request most likely does not fix the `contains` function if `Bw` wraps around `pi`. However, it does fix a major regression in `sample` that was causing issues with the `block_sorting` and `table_clearing` demos, so I am merging it.

I'll create another issue to track the `contains` issue.

psigen:
:+1: I think we need to carefully figure out a way to keep `Bw_cont` and `Bw` in sync and only use one for the checks.

mkoval:
I logged the `contains` issue in #199, so I am merging this.

PR #196, opened 2015-09-30
Added separate exception for trajectory precondition violations.

psigen:
This creates a new `prpy.exception` which handles the specific case where a trajectory cannot start execution due to a violation of a required precondition for execution, such as the robot start location.


jeking04:
I would like to make a thumbs up. But I do not know how.

psigen:
`:thumbsup:`

Issue #207, opened 2015-10-21
Use of `hasattr()` in `__getattr__()` overload of `prpy.base.robot.py` causes infinite recursion.


psigen:
The recent addition of `self.detector` support to prpy revealed the following bad logic in the `prpy.base.robot` attribute logic:

If a `prpy.base.robot` subclass does not have a `detector` attribute (or the `planner` or `actions` attributes are removed or missing for any reason), then all attribute lookups on a robot will hang forever in a recursion loop.

This is because  `__getattr__(self, anything)` internally calls `hasattr(self, 'detector')`, which calls `__getattr__(self, 'detector')`, which starts the following infinite loop:

```
/home/pkv/indigo/workspace/src/deps/planning/prpy/src/prpy/bind.pyc in intercept(self, name)
    112             # we explicitly invoke it on object.
    113             if hasattr(canonical_instance, '__getattr__'):
--> 114                 return canonical_instance.__getattr__(name)
    115             else:
    116                 raise

/home/pkv/indigo/workspace/src/deps/planning/prpy/src/prpy/base/robot.py in __getattr__(self, name)
    116                 return delegate_method(self, *args, **kw_args)
    117             return wrapper_method
--> 118         elif (hasattr(self, 'detector') and self.detector is not None
    119               and self.detector.has_perception_method(name)):
    120 

/home/pkv/indigo/workspace/src/deps/planning/prpy/src/prpy/bind.pyc in intercept(self, name)
    112             # we explicitly invoke it on object.
    113             if hasattr(canonical_instance, '__getattr__'):
--> 114                 return canonical_instance.__getattr__(name)
    115             else:
    116                 raise

/home/pkv/indigo/workspace/src/deps/planning/prpy/src/prpy/base/robot.py in __getattr__(self, name)
    116                 return delegate_method(self, *args, **kw_args)
    117             return wrapper_method
--> 118         elif (hasattr(self, 'detector') and self.detector is not None
    119               and self.detector.has_perception_method(name)):
    120 
```

psigen:
This is a fairly major bug, as currently `prpy.base.robot` does NOT include a `self.detector = None` in its init, as it does for `self.action` and `self.planner`.

Even if such an attribute exists, there should be a mechanism to not infinitely recurse if it is removed for some reason.  One method is to check the attribute name that is being queried in `__getattr__()` before allowing a `hasattr()` lookup: https://github.com/personalrobotics/prpy/pull/208.

PR #208, opened 2015-10-21
Check attribute name before doing hasattr().

psigen:
This addresses https://github.com/personalrobotics/prpy/issues/207 by first checking that the attribute that is being accessed is not named the same as the attribute that is being used to lookup other methods.


mkoval:
:+1: Thanks @psigen.

PR #209, opened 2015-10-21
add MoveUntilTouch

Shentheman:
This is used for the block sorting task in ada. Ada may use MoveUntilTouch to grab a block. I may need to modify it to test on the real robot. And I will remove the extra comments in the future.


mkoval:
We do not have the necessary logic implemented at the ros_control level to run `MoveUntilTouch` on the MICO. All of the functionality on MICO should work both in simulation and on the real robot, so I can't merge this until we have a solution that works on the real robot.

I am closing this pull request until then.

PR #210, opened 2015-10-23
Added helper functions to check if the robot is at a goal configuration

DavidB-CMU:
@mkoval 
IsAtActiveDOFConfiguration() is the original function I asked you about, but with a more descriptive name.

IsAtTrajectoryEnd() is a new function that you suggested, based on IsAtTrajectoryStart() which already exists in util.py

Both have been tested.

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -680,9 +680,9 @@ def FindCatkinResource(package, relative_path):
                       relative_path))
 
 
-def IsAtTrajectoryStart(robot, trajectory):
+def IsAtTrajectoryWaypoint(robot, trajectory, waypoint_idx):
     """
-    Check if robot is at the first waypoint of a trajectory.
+    Check if robot is at a particular waypoint in a trajectory.
 
     This function examines the current DOF values of the specified
     robot and compares these values to the first waypoint of the
@@ -690,12 +690,14 @@ def IsAtTrajectoryStart(robot, trajectory):
     differ by less than the DOF resolution of the specified joint/axis
     then it will return True. Otherwise, it returns False.
 
-    NOTE: This is used in ExecuteTrajectory()
+    NOTE: This is used in ExecuteTrajectory(),
+                            IsAtTrajectoryStart(), and
+                              IsAtTrajectoryEnd()
 
     @param robot: The robot whose active DOFs will be checked.
-    @param trajectory: The trajectory whose start configuration will
-                       be checked.
-    @returns: True The robot is at the end of the trajectory.
+    @param trajectory: The trajectory containing the waypoint
+                       to be checked.
+    @returns: True The robot is at the desired position.
               False One or more joints differ by DOF resolution.
     """
     if trajectory.GetNumWaypoints() == 0:
@@ -738,9 +740,6 @@ def IsAtTrajectoryStart(robot, trajectory):
             return False
 
     else:
-        # Get the FIRST waypoint
-        waypoint_idx = 0
-
         # Get joint indices used in the trajectory,
         # and the joint positions at this waypoint
         waypoint = trajectory.GetWaypoint(waypoint_idx)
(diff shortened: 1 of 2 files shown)


src/prpy/util.py
```diff
@@ -751,6 +751,88 @@ def IsAtTrajectoryStart(robot, trajectory):
...
+    @param trajectory: the trajectory whose end configuration will be checked
+    @returns: True if the robot's active DOFs match the given trajectory
+              False if one or more active DOFs differ by DOF resolution
+    """
+    cspec = trajectory.GetConfigurationSpecification()
+    needs_base = HasAffineDOFs(cspec)
+    needs_joints = HasJointDOFs(cspec)
+    
+    if needs_base and needs_joints:
+        raise ValueError('Trajectories with affine and joint DOFs are not supported in IsAtTrajectoryEnd()')
```
psigen: Wrap lines to 80 chars to be pep8.
:neckbeard:

src/prpy/util.py
```diff
@@ -751,6 +751,88 @@ def IsAtTrajectoryStart(robot, trajectory):
...
+
+    if trajectory.GetEnv() != robot.GetEnv():
+        raise ValueError('The environment attached to the trajectory does not match the environment attached to the robot in IsAtTrajectoryEnd()')
+
+    if needs_base:
+        raise ValueError('Trajectories with affine DOFs are not supported in IsAtTrajectoryEnd()')
+
+    else:
+        last_waypoint_idx = trajectory.GetNumWaypoints() - 1
```
mkoval: This will cause an `IndexError` on an empty trajectory. I see two options for how to handle this:

1. check for an empty trajectory and throw an error with a good error message
2. return `True`

@psigen Any preference?

src/prpy/util.py
```diff
@@ -751,6 +751,88 @@ def IsAtTrajectoryStart(robot, trajectory):
...
+    dof_indices = robot.GetActiveDOFIndices()
+
+    # Get current configuration of robot for used indices
+    with robot.GetEnv():
+        joint_values = robot.GetDOFValues(dof_indices)
+
+    # If any joint is not at the goal position, return False
+    for i in xrange(1, len(goal_config)):
+        delta_value = abs( joint_values[i] - goal_config[i] )
+        if delta_value > epsilon:
```
mkoval: This should use `SubtractValue` like above.

src/prpy/util.py
psigen: Well, semantically, I wouldn't say we _are_ at the empty trajectory endpoint, so I think a `ValueError` or `False` would make more sense.

mkoval:
I am a bit concerned that `IsAtActiveDOFConfiguration` and `IsAtTrajectoryEnd` have similar, but different, logic. I suggest that you refactor this as follows:
 
1. Refactor `IsAtActiveDOFConfiguration` to accept a target configuration, instead of a trajectory.
2. Write a new function `IsActiveDOFsAtEnd` that gets the last waypoint in the trajectory and calls (1).
2. Refactor `IsAtTrajectoryEnd` to set the DOFs inside the trajectory inside a `RobotStateSaver` and call (2).

@DavidB-CMU Does that seem reasonable?

DavidB-CMU:
I will open a new PR when these issues are addressed.

mkoval:
I'd prefer to continue using this pull request, so we have one coherent discussion thread, unless you have a strong reason to open a new one. Anything you commit to this branch will show up here when you push to Github.

DavidB-CMU:
@mkoval @psigen All the above issues have been addressed, as per a conversation with Michael.  
The commit diff is messy, so please review the entire block of code here:  
https://github.com/DavidB-CMU/prpy/blob/master/src/prpy/util.py#L683

DavidB-CMU:
@mkoval @psigen  I added some tests for these 3 functions.

mkoval:
It's troubling that `IsAtTrajectoryStart` handles this case, but `IsAtTrajectoryEnd` does not. These two functions are also nearly a straight copy-and-paste.

**The simplest option** is to add a `IsAtTrajectoryWaypoint` function that performs this check on an arbitrary waypoint index. Then, you implement the two existing functions as:
- `IsAtTrajectoryStart(robot, traj)` calls
```python
IsAtTrajectoryWaypoint(robot, traj, 0)
```
- `IsAtTrajectoryEnd` calls
```python
IsAtTrajectoryWaypoint(robot, traj, traj.GetNumWaypoints() - 1)
```

**Alternatively,** you could make the function take: (1) a robot, (2) a `ConfigurationSpecification`, and (3) a waypoint in that `ConfigurationSpecification`. Then, the implementation would look something like this:

- `IsAtTrajectoryStart(robot, traj)` calls
```python
IsAtTrajectoryWaypoint(robot, traj.GetConfigurationSpecification(), traj.GetWaypoint(0))
```
- `IsAtTrajectoryEnd` calls
```python
IsAtTrajectoryWaypoint(robot, traj.GetConfigurationSpecification(), traj.GetWaypoint(traj.GetNumWaypoints() - 1))
```

**I'd prefer** the second option, since it lets us call `IsAtTrajectoryWaypoint` on the return value of `Sample`. @DavidB-CMU What do you think?

Sorry for not suggesting this sooner. :sweat:

DavidB-CMU:
@mkoval Ok I committed your first idea to add IsAtTrajectoryWaypoint(). I agree that it is much more tidy.

mkoval:
Thanks! This looks good.

PR #215, opened 2015-10-28
added getmanipindex function to util

rachelholladay:
As requested in the discussion on the rogue merge in herbpy, I added a simple function to util that takes in a robot and returns the manip and manip_idx. These couple of lines were coming up a lot so the util function is to reduce code repetition.


mkoval:
The pull request @rachelholladay is https://github.com/personalrobotics/herbpy/pull/54.

:+1: Thanks.

Issue #226, opened 2015-11-10
Implement alternative to GetCollisionCheckPts() for checking piece-wise linear trajectories


DavidB-CMU:
Planners like SnapPlanner currently do collision-checking by calling GetCollisionCheckPts(), which samples the trajectory at DOF resolution by using bi-section to determine an appropriate amount of joint motion for each joint.  
This is not using the correct distance metric and we would also like to collision-check trajectories in a more optimal, non-linear order.

A new function GetLinearCollisionCheckPts() allows you to specify the distance metric (L1, L2, L_infinity) and the sampling order (linear, Van-der-Corput).

This is linear sampling, where the end of the trajectory is checked last:
[0.0, 0.25, 0.5, 0.75, 1.0]

This is sampling using binary van der Corput sequence, where the end-points are checked first, then the mid-point, and so on:
[0.0, 1.0, 0.5, 0.25, 0.75]

This other issue is somewhat related:
#204 (GetCollisionCheckPts() fails on a single-waypoint trajectory)

DavidB-CMU:
Pull request personalrobotics/prpy/pull/228

DavidB-CMU:
**Performance comparison:**

To move the arm from 'relaxed_home' to some pose in front of the robot:
**Sampling method:** linear
67 checks, 0.23 sec
**Sampling method:** binary Van der Corput
67 checks, 0.23 sec

With a Fuze bottle in-between the start and goal poses:
**Sampling method:** linear
24 checks, 0.1 sec, collision detected.
**Sampling method:** binary Van der Corput
2 checks, 0.03 sec, collision detected.

DavidB-CMU:
This was merged on 5 Dec 2015. Closing issue.

PR #228, opened 2015-11-11
New method for getting collision sample points, Switch SnapPlanner...

DavidB-CMU:
More details in issue #226

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -976,6 +976,52 @@ def ConvertIntToBinaryString(x, reverse=False):
     return ''.join(bin(x)[2:])
 
 
+def VanDerCorputSequence(lower=0.0, upper=1.0, include_endpoints=True):
+    """
+    Generate the binary Van der Corput sequence, where each value
+    is a dyadic fraction re-scaled to the desired range.
+
+    For example, on the interval [0,1], the first 5 values of
+    the Van der Corput sequence are:
+    [0.0, 1.0, 0.5, 0.5, 0.75]
+
+    @param float lower: The first value of the range of the sequence.
+    @param float upper: The last value of the range of the sequence.
+
+    @param bool include_endpoints: If True, the output sequence will
+                                   include the value 'lower' and the
+                                   value 'upper'.
+                                   If False, these endpoint values
+                                   will not be returned.
+
+    @returns generator: A sequence of float values.
+    """
+    from itertools import count, chain
+
+    if include_endpoints == True:
+        endpoints = (0.0, 1.0)
+    else:
+        endpoints = None
+
+    # Get a sequence of reversed binary numbers:
+    # '1', '01', '11', '001', '101', '011', '111', '0001', ....
+    #
+    # Note: count(1) is a generator, starting at 1, making steps of 1.
+    reverse_binary_seq = (ConvertIntToBinaryString(x, True) for x in count(1))
+
+    # From the reversed binary sequence, generate the Van der Corput
+    # sequence, for which:  0.0 < x < 1.0  (the end-points are excluded)
+    # 0.5, 0.25, 0.75, 0.125, 0.625, 0.375, 0.875, 0.0625, ....
+    #
+    # Note: int(x,2) converts the binary string (base 2) to an integer.
+    raw_seq = (float(int(x,2)) / (2**len(x)) for x in reverse_binary_seq)
+
+    # Scale the Van der Corput sequence across the desired range
+    # and optionally add the end-points.
+    scale = float(upper - lower)
+    return (scale * val + lower for val in chain(endpoints or [], raw_seq))
+
+
 def GetCollisionCheckPts(robot, traj, include_start=True, start_time=0.,
                          first_step=None, epsilon=1e-6):
     """
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -955,6 +955,27 @@ def CheckJointLimits(robot, q):
             description='position')
 
 
+def ConvertIntToBinaryString(x, reverse=False):
+    """
+    Convert an integer to a binary string.
+
+    Optionally reverse the output string, which is
+    required for producing a Van Der Corput sequence.
+
+    @param int  x:       The number to be converted.
+    @param bool reverse: If True, the output string will be reversed.
+
+    @returns string: A binary number as a string.
+    """
+    if type(x) != int:
+        raise ValueError('Input number must be an integer')
+
+    if reverse == True:
+        return ''.join(reversed(bin(x)[2:]))
+
+    return ''.join(bin(x)[2:])
+
+
 def GetCollisionCheckPts(robot, traj, include_start=True, start_time=0.,
                          first_step=None, epsilon=1e-6):
     """
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1043,6 +1043,11 @@ def SampleTimeGenerator(start, end, step=1):
 
     @returns generator: A sequence of float values.
     """
+    if end <= start:
+        raise ValueError("The 'end' value must be greater than "\
+                         "the 'start' value.")
+    if not (step > 0):
+        raise ValueError("The 'step' value must be positive.")
     t = start
     prev_t  = 0.0
     while t <= numpy.floor(end):
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1236,7 +1236,8 @@ def GetLinearCollisionCheckPts(robot, traj, norm_order=2, sampling_func=None):
 
     # Make sure trajectory is linear in joint space
     try:
-        # type can be 'linear' or 'quadratic'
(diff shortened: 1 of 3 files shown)


mkoval:
There appear to be some syntax issues in `util.py` (see the Travis build log).

src/prpy/util.py
```diff
@@ -986,6 +1184,140 @@ def GetCollisionCheckPts(robot, traj, include_start=True, start_time=0.,
...
+    if sampling_order == 'van_der_corput':
+        # An approximate Van der Corput sequence, between the
+        # start and end points
+        seq = VanDerCorputSampleGenerator(0, traj_duration, step=2)
+    elif sampling_order == 'linear':
+        # (default) Linear sequence, from start to end
+        seq = SampleTimeGenerator(0, traj_duration, step=2)
+    else:
+        error = "Unknown sampling_order '" + sampling_order + "' "
+        raise ValueError(error)
```
mkoval: I suggest that you change the `sampling_order` parameter to directly accept a generator function. The function would accept `(start, end, step)` and return a generator. It could default to  `GetLinearCollisionCheckPts` to replicate the current default behavior.

DavidB-CMU:
Attempting to force Travis to re-build...

src/prpy/util.py
```diff
@@ -914,6 +914,204 @@ def ComputeUnitTiming(robot, traj, env=None):
...
+
+    @param int  x:       The number to be converted.
+    @param bool reverse: If True, the output string will be reversed.
+
+    @returns string: A binary number as a string.
+    """
+    if type(x) != int:
+        raise ValueError('Input number must be an integer')
+
+    if reverse == True:
```
mkoval: `if reverse:` is more Pythonic.

src/prpy/util.py
```diff
@@ -914,6 +914,204 @@ def ComputeUnitTiming(robot, traj, env=None):
...
+                                   will not be returned.
+
+    @returns generator: A sequence of float values.
+    """
+    from itertools import count, chain
+
+    if include_endpoints == True:
+        endpoints = (0.0, 1.0)
+    else:
+        endpoints = None
```
mkoval: If you set `endpoints = []` here you can skip the `endpoints or []` check below.

src/prpy/util.py
```diff
@@ -914,6 +914,204 @@ def ComputeUnitTiming(robot, traj, env=None):
...
+
+    @param bool include_endpoints: If True, the output sequence will
+                                   include the value 'lower' and the
+                                   value 'upper'.
+                                   If False, these endpoint values
+                                   will not be returned.
+
+    @returns generator: A sequence of float values.
+    """
+    from itertools import count, chain
```
mkoval: Should check for `lower < upper`.

src/prpy/util.py
```diff
@@ -914,6 +914,204 @@ def ComputeUnitTiming(robot, traj, env=None):
...
+
+    For example, on the interval [0.0,5.0], the sequence is:
+    [0.0, 1.0, 2.0, 3.0, 4.0, 5.0]
+
+    @param float start: The start value of the sequence.
+    @param float end:   The last value of the sequence.
+    @param float step:  The step-size between values.
+
+    @returns generator: A sequence of float values.
+    """
```
mkoval: Also check that `start < end` here. On a side note, it would be nice to use the same names for these arguments in the two functions so they are interchangeable even if you are pass keyword arguments.

mkoval: Nevermind on the second note. I now see that `SampleTimeGenerator` and `VanDerCorputSampleGenerator` are the two parallel functions and they already have identical argument names.

src/prpy/util.py
```diff
@@ -986,6 +1184,140 @@ def GetCollisionCheckPts(robot, traj, include_start=True, start_time=0.,
...
+
+    # If trajectory is already timed, strip the deltatime values
+    if IsTimedTrajectory(traj):
+        traj = UntimeTrajectory(traj)
+
+    traj_cspec = traj.GetConfigurationSpecification()
+
+    # Make sure trajectory is linear in joint space
+    try:
+        # type can be 'linear' or 'quadratic'
```
mkoval: This comment is not accurate: this function only works on `linear` interpolation. There are many other values of `interpolation`, too, if you are using higher-order splines.

src/prpy/util.py
```diff
@@ -986,6 +1184,140 @@ def GetCollisionCheckPts(robot, traj, include_start=True, start_time=0.,
...
+        interp_type = traj_cspec.GetGroupFromName('joint_values').interpolation
+    except openravepy.openrave_exception:
+        raise ValueError('Failed calling GetGroupFromName()')
```
mkoval: A better error message would be something like "Trajectory does not have a `joint_values` group."

tests/test_util.py
```diff
@@ -142,8 +158,444 @@ def test_IsAtTrajectoryEnd_ReturnsTrue(self):
...
+            prpy.util.CheckJointLimits(self.robot, q0)
+        except JointLimitError:
+            pass # test passed
+        except Exception, e:
+            error = 'Unexpected exception thrown: ' + str(e.message)
+            self.fail(error)
+        else:
+            self.fail('Expected exception not thrown')
+
+    def test_CheckJointLimits_MaxLimitAll(self):
```
mkoval: Please add a `MaxLimitOneJoint` test, too.

tests/test_util.py
```diff
@@ -142,8 +158,444 @@ def test_IsAtTrajectoryEnd_ReturnsTrue(self):
...
+        q0 = [0.0, 0.0, 0.0, -1.0, 0.0, 0.0, 0.0]
+        try:
+            prpy.util.CheckJointLimits(self.robot, q0)
+        except JointLimitError:
+            pass # test passed
+        except Exception, e:
+            error = 'Unexpected exception thrown: ' + str(e.message)
+            self.fail(error)
+        else:
+            self.fail('Expected exception not thrown')
```
mkoval: The `unittest` module has built-in functionality for doing this:
```python
with self.assertRaises(JointLimitError):
    prpy.util.CheckJointLimits(self.robot, q0)
```

This comment applies to a bunch of other tests, too.

tests/test_util.py
```diff
@@ -142,8 +158,444 @@ def test_IsAtTrajectoryEnd_ReturnsTrue(self):
...
+
+    def test_CheckJointLimits_ZeroPosition(self):
```
mkoval: A better name for this would be `WithinLimits_DoesNotThrow` or something like that. There's no reason to think that the zero configuration is within limits (e.g. it's not on HERB).

tests/test_util.py
```diff
@@ -142,8 +158,444 @@ def test_IsAtTrajectoryEnd_ReturnsTrue(self):
...
+        desired_num_checks = 1
+        traj = self.CreateTrajectory(q0, q0) # makes traj with 1 waypoint
+        checks = prpy.util.GetLinearCollisionCheckPts(self.robot, \
+                                                      traj, \
+                                                      norm_order=2, \
+                                                      sampling_order='linear')
+        num_checks = sum(1 for x in checks)
+        if num_checks != desired_num_checks:
+            error = str(num_checks) + ' is the wrong number of check pts.'
+            self.fail(error)
```
mkoval:
```python
self.assertEqual(num_checks, desired_num_checks)
```

tests/test_util.py
```diff
@@ -142,8 +158,444 @@ def test_IsAtTrajectoryEnd_ReturnsTrue(self):
...
+
+
+    # ConvertIntToBinaryString()
+
+    def test_ConvertIntToBinaryString(self):
+        # The integer 6 in binary is "110"
+        binary_string = prpy.util.ConvertIntToBinaryString(6)
+        expected_string = '110'
+        error = 'The binary value ' + binary_string + \
+                ' does not equal the expected value ' + expected_string
```
mkoval: Nitpick: It would be more Pythonic to do:
```python
'The binary value {:s} does not equal the expected value {:s}'.format(binary_string, expected_string)
```

mkoval:
This looks **really** good. Awesome work, @DavidB-CMU!

My only comment on the code itself is that I'd prefer to take the sequence generator as an argument, instead of specifying a string from a hard-coded set. I think this should be a relatively striaghtforward change (and may even eliminate some code).

The rest of the comments are minor nitpicks about the unit tests. In general, I'd prefer if you used the built-in testing methods (like `assertRaises` and `assertEqual`) instead of implementing the logic yourself with `fail`. The built-in methods produce nice (and consistent) output when the assertion fails.

siddhss5:
I had the same comment: much cleaner to make the sequence generator an argument. Python is really fantastic for exactly such things!

psigen:
Is this branch still happening?

DavidB-CMU:
@psigen Yes this pull-request will be finished soon :-)

DavidB-CMU:
Please review this Pull Request.

mkoval:
@DavidB-CMU Thanks for making that change!

PR #236, opened 2015-11-25
Fix chomp OptimizeTrajectory to actually return the optimized trajectory (not the original)

cdellin:


mkoval:
Good catch! :smile:

PR #237, opened 2015-11-30
Feature/perception simtrack

mklingen:
Added a perception module for `simtrack`. This just runs the simtrack detector and returns objects. Only supports fuze bottle and poptarts right now. `simtrack` node has to be running continuously on `herb2`, which must be plugged in to a monitor.

### DIFF
--- a/src/prpy/perception/__init__.py
+++ b/src/prpy/perception/__init__.py
@@ -32,3 +32,4 @@ from base import PerceptionModule, PerceptionMethod
 from apriltags import ApriltagsModule
 from simulated import SimulatedPerceptionModule
 from rock_module import RockModule
+from simtrack import SimtrackModule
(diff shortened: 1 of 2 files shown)


mkoval:
This pull request currently has a file level `import simtrack_msgs` that breaks if that package is not installed. Could you move this `import` into the function where it is used so it is only required when we are using SimTrack?

mklingen:
@mkoval okay.

mkoval:
Same thing with `tf` (see the build log).

mklingen:
Huh, I wonder why it builds with `vncc` detector then, since that file has all the same imports.

mklingen:
Oh, `vncc` isn't imported in `__init__.py`. I wonder if @Shushman intended that.

Shushman:
Me and @jeking04 used the vncc module at the time of the last Toyota demo and then we moved it to prpy during a cleanup but I did not get around to doing this integration.

PR #240, opened 2015-12-03
Added missing robot state saver to GreedyIK.

psigen:
This fixes a bug introduced by the exception handling PR where `collision_error` report generation accidentally changed the pose of the robot inside the planning loop.


mkoval:
Good catch. I discovered #245 in the process of reviewing this.

PR #246, opened 2015-12-14
Feature/tsr readme

jeking04:
Adding readme with TSR tutorial to prpy/src/prpy/tsr package.


mkoval:
I made a few small tweaks. Otherwise, it looks great!

PR #247, opened 2015-12-14
PlanWorkspacePath detects collision error only when the whole method …

lgw903:
…terminates with an error.


lgw903:
@mkoval  I think this will fix the issue you mentioned. Now it runs FindIKSolutions only when it's about to raise error for the whole plan.

mkoval:
Thanks!

PR #250, opened 2015-12-16
Adds futures to prpy.

psigen:
This backports the futures implementation from adapy upstream into prpy.

It does not use the futures for anything, but makes them available for use in later PRs and dependent libraries as part of the `prpy.futures` package.


mkoval:
Thanks for cleaning up that PR!

@Stefanos19, can you remove the duplicate code from AdaPy and use this instead?

PR #251, opened 2016-01-07
Add convenience function to compute forward kinematics

DavidB-CMU:
I can add a simple unit test that checks if a specific configuration returns the expected end effector transform?

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1027,16 +1027,21 @@ def CheckJointLimits(robot, q):
             description='position')
 
 
-def GetForwardKinematics(robot, q, manipulator=None):
+def GetForwardKinematics(robot, q, manipulator=None, frame=None):
     """
-    Get the forward kinematics for a specific joint configuration.
+    Get the forward kinematics for a specific joint configuration,
+    relative to the OpenRAVE world frame.
 
     @param openravepy.robot robot: The robot object.
     @param list             q:     List or array of joint positions.
-
+    @param manipulator
+    @param string frame: Get the end effector transform relative to a
+                         specific frame e.g. '/right/wam_base'
     @returns T_ee: The pose of the end effector (or last link in the
                    serial chain) as a 4x4 matrix.
     """
+    from tf import transformations
+
     if manipulator == None:
         manipulator = robot.GetActiveManipulator()
 
@@ -1050,6 +1055,10 @@ def GetForwardKinematics(robot, q, manipulator=None):
         robot.SetActiveDOFValues(q)
         T_ee = manipulator.GetEndEffectorTransform()
     # Robot state is restored
+
+    if frame != None:
+        T_ref_frame = robot.GetLink(frame).GetTransform()
+        T_ee = transformations.concatenate_matrices(numpy.linalg.inv(T_ref_frame), T_ee)
 
     return T_ee
 
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1027,6 +1027,27 @@ def CheckJointLimits(robot, q):
             description='position')
 
 
+def GetForwardKinematics(robot, q):
+    """
+    Get the forward kinematics for a specific joint configuration.
+
+    @param openravepy.robot robot: The robot object.
+    @param list             q:     List or array of joint positions.
+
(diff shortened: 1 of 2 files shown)


src/prpy/util.py
```diff
@@ -959,6 +959,27 @@ def CheckJointLimits(robot, q):
...
+    @param list             q:     List or array of joint positions.
+
+    @returns T_ee: The pose of the end effector (or last link in the
+                   serial chain) as a 4x4 matrix.
+    """
+    T_ee = None
+
+    # Save the robot state
+    sp = openravepy.Robot.SaveParameters
+    with robot.CreateRobotStateSaver():
```
mkoval: It would be better to only pass the `LinkTransformations` option here. Otherwise we'll save a bunch of extra state that we don't modify (e.g. joint limits, active DOFs, etc).

src/prpy/util.py
```diff
@@ -959,6 +959,27 @@ def CheckJointLimits(robot, q):
...
+        robot.SetActiveDOFValues(q)
+        T_ee = robot.GetActiveManipulator().GetTransform()
```
mkoval: I'd prefer if you took in `manipulator` as an argument - we're trying to move away from using `ActiveManipulator` as much as possible. You could make `manipulator` default to `GetActiveManipulator` to support the current API.

mkoval: Also, this likely should return `GetEndEffectorTransform`, not `GetTransform`. The former is equal to the latter offset by `GetLocalToolTransform`.

DavidB-CMU:
GetForwardKinematics() returns the end-effector transform w.r.t. the "/world" frame.

I'm thinking about adding an optional argument to specify the desired reference frame, 
for example to get the transform w.r.t. the base link of the WAM manipulator: 
```python
...

# Example:  frame = '/right/wam_base'
from tf import transformations
if frame != None:
    T_ref_frame = robot.GetLink(frame).GetTransform()
    T_ee = transformations.concatenate_matrices(numpy.linalg.inv(T_ref_frame), T_ee)
```

mkoval:
I noticed that a more general version of this functionality exists in `BodyPointsStatesFromJointState`. That code optionally supports velocities (`derivatives=[0,1]`) and accelerations (`derivatives=[0,1,2]`). It also allows you to query the state of any point, instead of only the end-effector frame. See #166 for some discussion about this API.

Is this sufficient for your needs? Or, if not, it would be nice to implement the new function in terms of the existing ones. Thoughts?

DavidB-CMU:
The GetForwardKinematics() function has 4 lines of code and is very easy to understand. I think we should add it.

I don't think GetForwardKinematics() should wrap a complicated function of 53 lines.

src/prpy/util.py
```diff
@@ -959,6 +959,42 @@ def CheckJointLimits(robot, q):
...
+    robot_saver = robot.CreateRobotStateSaver(sp.LinkTransformation)
+
+    with robot_saver:
+        robot.SetActiveDOFValues(q)
+        T_ee = manipulator.GetEndEffectorTransform()
+    # Robot state is restored
+
+    if frame != None:
+        T_ref_frame = robot.GetLink(frame).GetTransform()
+        T_ee = transformations.concatenate_matrices(numpy.linalg.inv(T_ref_frame), T_ee)
```
mkoval: I'd prefer to not add a `tf` dependency for this. Could you use `numpy.dot` instead?

src/prpy/util.py
```diff
@@ -959,6 +959,42 @@ def CheckJointLimits(robot, q):
             description='position')
 
 
+def GetForwardKinematics(robot, q, manipulator=None, frame=None):
```
mkoval: An API suggestion: remove the `robot` parameter and just take `manipulator`. You can get the robot from `manipulator.GetRobot()`. Thoughts?

mkoval:
I am fine with merging this if you: (1) remove the `tf` dependency, see above, and (2) fix the failed build on Travis.

DavidB-CMU:
- Removed `tf` dependency 
 - Added an extra unit test 
 - Rebased the commits to make Travis happy

PR #253, opened 2016-01-10
Adding Util function for RoGuE

rachelholladay:
As @mkoval mentioned in personalrobotics/ada#51 it makes sense to pull out this transforming as a util function. 

Once merge I'll push the slight change across the robots

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1599,16 +1599,21 @@ def GetManipulatorIndex(robot, manip=None):
 
     return (manip, manip_idx)
 
-def GetPointFrom(env, focus):
+def GetPointFrom(focus):
     """
     Given a kinbody, array or transform, returns the xyz
     location. 
-    param env The environment where the item exists
-    param focus THe area to be referred to
+    param focus The area to be referred to
     """
-    #Pointing at an object
-    if isinstance(focus, (openravepy.KinBody, openravepy.KinBody.Link)):
-        with env:
+    #Pointing at a kinbody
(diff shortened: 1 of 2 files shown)


src/prpy/util.py
```diff
@@ -1598,3 +1598,35 @@ def GetManipulatorIndex(robot, manip=None):
...
+def GetPointFrom(env, focus):
+    """
+    Given a kinbody, array or transform, returns the xyz
+    location. 
+    param env The environment where the item exists
+    param focus THe area to be referred to
+    """
+    #Pointing at an object
+    if isinstance(focus, (openravepy.KinBody, openravepy.KinBody.Link)):
+        with env:
```
mkoval: You could replace this with `focus.GetEnv()` and remove the `env ` parameter entirely.

mkoval: Actually, that's not entirely true. It would be `focus.GetEnv()` for `KinBody` and `focus.GetParent().GetEnv()` for `Link`. But my point still stands.

rachelholladay: Done.

mkoval:
Can you write a few unit tests for this function? I think it would be sufficient to have one test per type of object supported (e.g. `KinBody`, `Link`, etc.) and one test to verify that unsupported types throw an exception.

rachelholladay:
@mkoval Tests added.

mkoval:
:+1: Thanks for making those changes @rachelholladay.

I would have merged this sooner, but I didn't notice that you finished the unit test refactor.

PR #254, opened 2016-01-10
Fixed issue with CloseHand() for ADA in sim 

Stefanos19:


mkoval:
Note that `MoveHand` function ignores collisions and will always terminate with the hand fully closed, even if it is penetrating an object. The `robot.task_manipulation.CloseFingers()` function simulates closing the fingers until they are in collision. This is why we have separate functions for `CloseFingers` and `OpenFingers`.

mkoval:
@Stefanos19 Can you take a look at my comment above?

Stefanos19:
I think the issue is that the command:  robot.task_manipulation.prob.SendCommand('CloseFingers') returns the following error: 
[grasperplanner.cpp:258 PlanPath] no objects in environment

mkoval:
I will merge this for now. However, it is really not correct - we should figure out why `CloseFinger`s is failing.

@Stefanos19 Please make an issue about this.

PR #258, opened 2016-01-19
Implement Servo() on MicoArm

sjavdani:
Implemented joint velocity commands through the Servo function for the mico arm. Also added a watchdog that sends zero velocity after 0.25 seconds of no commands sent to the arm.

Not included is the code to activate all the controllers - I'll add that directly to adapy, since it uses the controller interface included there.

### DIFF
--- a/src/prpy/base/mico.py
+++ b/src/prpy/base/mico.py
@@ -32,6 +32,8 @@ import openravepy
 from manipulator import Manipulator
 from std_msgs.msg import Float64
 import rospy
+from ..util import Watchdog
+from functools import partial
 
 class Mico(Manipulator):
     def __init__(self, sim,
@@ -69,11 +71,13 @@ class Mico(Manipulator):
                 self.GetName(), '', self.GetArmIndices(), 0, True)
             self.servo_simulator = ServoSimulator(self, rate=20,
                                                   watchdog_timeout=0.1)
-            
         else:
-        #if not simulation, create publishers for each joint
+            #if not simulation, create publishers for each joint
             self.velocity_topic_names = ['vel_j'+str(i)+'_controller/command' for i in range(1,7)]
-            self.velocity_publishers = [rospy.Publisher(topic_name, Float64) for topic_name in self.velocity_topic_names]
+            self.velocity_publishers = [rospy.Publisher(topic_name, Float64, queue_size=1) for topic_name in self.velocity_topic_names]
+            
+            #create watchdog to send zero velocity
+            self.servo_watchdog = Watchdog(timeout_duration=0.3, handler=lambda: self.SendVelocitiesToMico([0.,0.,0.,0.,0.,0.,]))
 
     def CloneBindings(self, parent):
         super(Mico, self).CloneBindings(parent)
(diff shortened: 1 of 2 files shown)


src/prpy/base/mico.py
```diff
@@ -99,4 +109,11 @@ def Servo(self, velocities):
             self.GetRobot().GetController().Reset(0)
             self.servo_simulator.SetVelocity(velocities)
         else:
-            raise NotImplementedError('Servo is not implemented.') 
+            self.SendVelocitiesToMico(velocities)
+            #reset watchdog timer
+            self.servo_watchdog.reset()
+
+
+    def SendVelocitiesToMico(self, velocities):
```
mkoval: I'd prefer to put this in `Servo`, since this code has a subset of that functionality: it doesn't work in simulation. Is there a strong reason to keep it as a separate function?

src/prpy/util.py
```diff
@@ -497,6 +497,21 @@ def stop(self):
...
+        self.timer.start()
+
+    def reset(self):
+        self.timer.cancel()
+        self.timer = threading.Timer(self.timeout_duration, self.handler)
+        self.timer.start()
+        
+    def stop(self):
+        self.timer.cancel()
+
```
mkoval: Please add docstrings to this class.

src/prpy/util.py
```diff
@@ -497,6 +497,21 @@ def stop(self):
...
+    def __init__(self, timeout_duration, handler):
+        self.timeout_duration = timeout_duration
+        self.handler = handler
+        self.timer = threading.Timer(self.timeout_duration, self.handler)
+        self.timer.start()
+
+    def reset(self):
+        self.timer.cancel()
+        self.timer = threading.Timer(self.timeout_duration, self.handler)
+        self.timer.start()
```
mkoval: I am a bit concerned about creating a new `Timer` every time we send a velocity command. My hunch is that every `Timer` starts a new thread. This will create and destroy a huge number of threads. There may be some gotchas about this, e.g. thread exhaustion, performance implications, but I am not positive. Maybe @cdellin or @psigen is more informed?

If this indeed an issue, then it is relatively straightforward to use one long-running thread per `Watchdog` by setting the expiration timestamp as a member variable.

src/prpy/base/mico.py
```diff
@@ -99,4 +109,11 @@ def Servo(self, velocities):
...
+        for velocity_publisher,velocity in zip(self.velocity_publishers, velocities):
+            velocity_publisher.publish(velocity)
```
mkoval: You should introduce a `Lock` around this loop. This function is called from the user's thread(s) and the watchdog thread. I am not sure if `publish` is threadsafe. Even if it is, bad things will happen if `publish` commands are interleaved between two threads.

mkoval:
The high-level implementation looks good. I posted a few comments related to threading on the commits.

src/prpy/base/mico.py
sjavdani: This is so the watchdog can call something without going through Servo. If I had it in servo, and the watchdog called Servo, then it would reset itself

sjavdani:
Ok, added lock, added docstring and modified the use of timer. Comment about servo vs. set velocity inline. Now, it will reset it's internal timer until the handler is called once, at which time the thread will exit. if you call reset again, and the thread has exited, it will spawn a new thread.

src/prpy/base/mico.py
mkoval: Good point! :+1:

PR #259, opened 2016-01-22
Added a helper function to extract joint groups.

psigen:
This commit creates a helper function that copies OpenRAVE's
hardcoded table of joint derivative names and adds an accessor
function for retrieving them from a configuration specification.

This allows users to directly query the n-th derivative group of
a trajectory without having to iterate through other derivatives.

I could not think of a cleaner way to do this, but at least this
avoid having the values hard-coded in multiple places.

Note that this function only handles joint value derivatives.  A
corresponding function for affine derivatives could be added in the
future.

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1374,6 +1374,39 @@ def IsInCollision(traj, robot, selfcoll_only=False):
     return False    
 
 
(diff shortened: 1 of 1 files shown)


mkoval:
It's sad that this is necessary, but it is a very useful function to have.

Issue #264, opened 2016-02-07
Add PlanWorkspacePath() method to VectorFieldPlanner


dbworth:
New feature, this will add a PlanWorkspacePath() method to Vector Field Planner.

Pull request #265

DavidB-CMU:
Function has been added in PR https://github.com/personalrobotics/prpy/pull/269

PR #265, opened 2016-02-07
(moved) Add PlanWorkspacePath() method to VectorFieldPlanner

dbworth:
This PR is to add a PlanWorkspacePath() method to VectorFieldPlanner.

Issue #264 

.

Questions about VectorFieldPlanner in general:
 - If you call PlanToEndEffectorOffset() for some distance, the planner never reaches that distance. Can we fix this?
 - Should we rename the "max_distance" argument?  it seems confusing to me.

Questions about this PR:  
 - Potentially PlanToEndEffectorOffset() could be refactored to call PlanWorkspacePath, however I think PlanToEndEffectorOffset() has a more efficient implementation.
 - Do all the PlanWorkspacePath() methods in each planner need to have the same prototype?
 - To implement PlanWorkspacePath(), I call GetMinDistance... in both of the callbacks, so maybe this is not optimal.

### DIFF
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -243,8 +243,7 @@ class VectorFieldPlanner(BasePlanner):
         from .exceptions import (
             CollisionPlanningError,
             SelfCollisionPlanningError,
-            TimeoutPlanningError,
-            JointLimitError
+            TimeoutPlanningError
         )
         from openravepy import CollisionReport, RaveCreateTrajectory
         from ..util import ComputeJointVelocityFromTwist, GetCollisionCheckPts, ComputeUnitTiming
@@ -263,8 +262,7 @@ class VectorFieldPlanner(BasePlanner):
         env = robot.GetEnv()
         active_indices = robot.GetActiveDOFIndices()
 
-        # Get the robot's joint limits
-        q_limit_min, q_limit_max = robot.GetActiveDOFLimits()
+        # Get the robot's joint velocity limits
         qdot_limit = robot.GetDOFVelocityLimits(active_indices)
 
         # Create a new trajectory matching the current
@@ -295,25 +293,9 @@ class VectorFieldPlanner(BasePlanner):
             if time.time() - time_start >= timelimit:
                 raise TimeLimitError()
 
-            # Check joint position limits. Do this before setting the DOF so we
-            # don't set the DOFs out of limits.
-            lower_position_violations = (q < q_limit_min)
-            if lower_position_violations.any():
-                index = lower_position_violations.nonzero()[0][0]
-                raise JointLimitError(robot,
-                    dof_index=active_indices[index],
-                    dof_value=q[index],
-                    dof_limit=q_limit_min[index],
-                    description='position')
-
-            upper_position_violations = (q > q_limit_max)
-            if upper_position_violations.any():
-                index = upper_position_violations.nonzero()[0][0]
-                raise JointLimitError(robot,
-                    dof_index=active_indices[index],
-                    dof_value=q[index],
-                    dof_limit=q_limit_max[index],
-                    description='position')
+            # Check joint position limits.
+            # We do this before setting the joint angles.
+            util.CheckJointLimits(robot, q)
 
             robot.SetActiveDOFValues(q)
 
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -113,7 +113,8 @@ class VectorFieldPlanner(BasePlanner):
 
             Succeed if end-effector pose is close to the goal pose.
             """
-            pose_error = util.GeodesicDistance(manip.GetEndEffectorTransform(),
+            pose_error = util.GetGeodesicDistanceBetweenTransforms(
+                                               manip.GetEndEffectorTransform(),
                                                goal_pose)
             if pose_error < pose_error_tol:
                 return Status.TERMINATE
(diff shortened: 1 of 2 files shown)


src/prpy/planning/vectorfield.py
```diff
@@ -95,8 +95,11 @@ def PlanToEndEffectorPose(self, robot, goal_pose, timelimit=5.0,
         manip = robot.GetActiveManipulator()
 
         def vf_geodesic():
+            """
+            Function defining a joint-space vector field.
+            """
```
mkoval: This `__docstring__` is misleading. An *arbitrary* vector field function implements a joint-space vector field. This function specifically implements a vector field that moves along the geodesic towards `goal_pose`.

I also don't think this function strictly needs a `__doscstring__` - although I'm certainly not going to argue with adding one - since it is only accessible inside this function.

src/prpy/planning/vectorfield.py
```diff
@@ -105,15 +108,23 @@ def vf_geodesic():
...
 
         def CloseEnough():
-            pose_error = util.GeodesicDistance(
-                        manip.GetEndEffectorTransform(),
-                        goal_pose)
+            """
+            Function defining the termination condition.
+
+            Succeed if end-effector pose is close to the goal pose.
+            """
```
mkoval: See my comment above on `vf_geodesic`. The same applies here.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
 
             return Status.CONTINUE
 
+        integration_timelimit = 10.0
```
mkoval: I believe this should use the `timelimit` argument.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+            """
+            Function defining a joint-space vector field.
+            """
+            T_ee_actual = manip.GetEndEffectorTransform()
+
+            # Find where we are on the goal trajectory by finding
+            # the the closest point
+            (_,t) = util.GetMinDistanceBetweenTransformAndWorkspaceTraj(
+                                                                   T_ee_actual,
+                                                                   traj)
```
mkoval: I believe this will fail if the trajectory has a loop (or near-loop) in it. You may want to mention this in the `__docstring__`.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+            raise ValueError('Position tolerance must be non-negative.')
+        elif angular_tolerance < 0.0:
+            raise ValueError('Angular tolerance must be non-negative.')
+
+        manip = robot.GetActiveManipulator()
+        Tstart = manip.GetEndEffectorTransform()
+
+        # Get the final end-effector pose
+        duration = traj.GetDuration()
+        T_ee_goal = openravepy.matrixFromPose(traj.Sample(duration)[0:7])
```
mkoval: This assumes that the trajectory has a single `AffineDOFs` group that is parameterized by a translation and a quaternion. If you are making this assumption, then you should check for this when you validate the input arguments (and add a few unit tests).

mkoval: Also, if this is a path (i.e. untimed), how are you `Sample`ing with a time?

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+            # Find where we are on the goal trajectory by finding
+            # the the closest point
+            (_,t) = util.GetMinDistanceBetweenTransformAndWorkspaceTraj(
+                                                                   T_ee_actual,
+                                                                   traj)
+            # Get the desired end-effector transform from
+            # the goal trajectory
+            t_step = 0.2
+            desired_T_ee = openravepy.matrixFromPose(traj.Sample(t)[0:7])
+            pose_ee_next = traj.Sample(t+t_step)[0:7]
```
mkoval: It looks like you are using `t_step` to approximate a finite-difference velocity. This is problematic because:

1. the trajectory is un-timed (see above)
2. you have no guarantee on the scale of the parameterization - 0.2 may be very large or very small

Ideally, we would directly extract the derivative from the trajectory and use this as the feed-forward term in the controller. If that is not possible, then we should at least convert the trajectory to a known parametrization (e.g. unit velocity) so we know the scale of `t_step`.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+
+            K_p = 0.5
+            direction = (1.0-K_p)*direction_perpendicular + \
+                         K_p*direction_parallel
+            # Normalize the direction vector
+            direction = numpy.array(direction, dtype='float')
+            direction /= numpy.linalg.norm(direction)
+
+            # Modify the twist
+            twist[0:3] = direction
```
mkoval: I'm not sure whats going on here. I think the current implementation will allow arbitrary drift of orientation, since the `(1 - Kp) * direction_parallel` feedback term is only applied to the translational component.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+
+            # Calculate joint velocities using an optimized jacobian
+            dqout, _ = util.ComputeJointVelocityFromTwist(robot, twist,
+                                              joint_velocity_limits=numpy.PINF)
```
mkoval: Why are you passing infinite velocity limits here? This is concerning for a few reasons:

1. Isn't `joint_velocity_limits` typically a vector of length `len(robot.GetActiveDOFs())`, not a scalar?
2. Does this function behave reasonably when the value is set to infinity?
3. Won't this result in very large steps?

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+
+            # Check if we have reached the end of the goal trajectory
+            dist_to_goal = util.GetGeodesicDistanceBetweenTransforms(T_ee_curr,
+                                                                     T_ee_goal)
+            if numpy.fabs(dist_to_goal) <= goal_pose_tol:
+                return Status.CACHE_AND_TERMINATE
+
+            return Status.CONTINUE
+
+        integration_timelimit = 10.0
```
mkoval: This should be an argument called `timelimit`.

src/prpy/planning/vectorfield.py
```diff
@@ -240,45 +393,41 @@ def FollowVectorField(self, robot, fn_vectorfield, fn_terminate,
...
+            """
+            # Set the joint values, without checking the joint limits
             robot.SetActiveDOFValues(q, CheckLimitsAction.Nothing)
+
             return fn_vectorfield()
 
         def fn_status_callback(t, q):
+            """
+            This is called for each collision check.
+            """
```
mkoval: This is overly specific. More than just collision checking happens in here - e.g. constraint violation checks.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +217,165 @@ def TerminateMove():
...
+                          timelimit=5.0,
+                          **kw_args):
         """
         Follow a joint space vectorfield to termination.
 
         @param robot
         @param fn_vectorfield a vectorfield of joint velocities
         @param fn_terminate custom termination condition
+        @param integration_timelimit The initial time step to be taken
+                                     by the integrator.
```
mkoval: This `__docstring__` is misleading. The initial step size is passed as the `first_step` parameter to the integrator and is currently hard-coded to 0.1. This argument appears to be the value of `t` for which we're trying to compute the value of the integral.

src/prpy/util.py
```diff
@@ -725,6 +725,58 @@ def GeodesicDistance(t1, t2, r=1.0):
...
+    """
+    Wrapper, to match GetGeodesicDistanceBetweenQuaternions()
+    """
+    return GeodesicDistance(T0, T1, r=1.0)
+
+
+def GetEuclideanDistanceBetweenTransforms(T0, T1):
+    """
+    Calculate the Euclidean distance between two 4x4 transforms.
+    (also called L2 or Pythagorean distance)
```
mkoval: This appears to only consider *translational* distance. This is not clear from the name of the function of the `__docstring__`.

src/prpy/util.py
```diff
@@ -725,6 +725,58 @@ def GeodesicDistance(t1, t2, r=1.0):
...
+    t_loc = 0.0
+    t = 0.0
+    while t < traj.GetDuration():
+        T_curr = openravepy.matrixFromPose(traj.Sample(t)[0:7])
+        error = GetEuclideanDistanceBetweenTransforms(T, T_curr)
+        if error < min_dist:
+            min_dist = error
+            t_loc = t
+        t = t + dt
+    return (min_dist, t_loc)
```
mkoval: You may want to add a special case that explicitly checks `traj.GetDuration()`. Otherwise, I could imagine this messing up the termination condition above.

mkoval: It would also be useful to return `T_curr` in the tuple so the user doesn't have to evaluate the trajectory again.

mkoval:
This is a very nice addition. I left a bunch of comments on the diff and have answered your questions below:

> If you call PlanToEndEffectorOffset() for some distance, the planner never reaches that distance. Can we fix this?

Can you be more specific?

> Should we rename the "max_distance" argument? it seems confusing to me.

What do you suggest as a new name? I am open to renaming this as long as:

1. we add a stub that maintains backwards compatibility; i.e. forward the `max_distance` parameter and raise a deprecation warning
2. update any other `PlanToEndEffectorOffset` methods that take this argument in the same way

> Potentially PlanToEndEffectorOffset() could be refactored to call PlanWorkspacePath, however I think PlanToEndEffectorOffset() has a more efficient implementation.
>
> To implement PlanWorkspacePath(), I call GetMinDistance... in both of the callbacks, so maybe this is not optimal.

That would be a nice change as long as, like you said, it does not have major performance implications. Can you time the two implementations and see if there's a noticeable performance difference?

> Do all the PlanWorkspacePath() methods in each planner need to have the same prototype?

The same arguments get passed to all planners in a `Sequence`. At a minimum, all of the planners should have identical *required arguments*. It's fine to have different optional arguments, but we do try to make them overlap as much as possible (e.g. we consistently use `timelimit` for all planners that have a planning time limit).

DavidB-CMU:
Moved to a new PR.

PR #269, opened 2016-02-12
Add PlanWorkspacePath() method to VectorFieldPlanner

DavidB-CMU:
This PR is to add a PlanWorkspacePath() method to VectorFieldPlanner.

### DIFF
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -243,8 +243,7 @@ class VectorFieldPlanner(BasePlanner):
         from .exceptions import (
             CollisionPlanningError,
             SelfCollisionPlanningError,
-            TimeoutPlanningError,
-            JointLimitError
+            TimeoutPlanningError
         )
         from openravepy import CollisionReport, RaveCreateTrajectory
         from ..util import ComputeJointVelocityFromTwist, GetCollisionCheckPts, ComputeUnitTiming
@@ -263,8 +262,7 @@ class VectorFieldPlanner(BasePlanner):
         env = robot.GetEnv()
         active_indices = robot.GetActiveDOFIndices()
 
-        # Get the robot's joint limits
-        q_limit_min, q_limit_max = robot.GetActiveDOFLimits()
+        # Get the robot's joint velocity limits
         qdot_limit = robot.GetDOFVelocityLimits(active_indices)
 
         # Create a new trajectory matching the current
@@ -295,25 +293,9 @@ class VectorFieldPlanner(BasePlanner):
             if time.time() - time_start >= timelimit:
                 raise TimeLimitError()
 
-            # Check joint position limits. Do this before setting the DOF so we
-            # don't set the DOFs out of limits.
-            lower_position_violations = (q < q_limit_min)
-            if lower_position_violations.any():
-                index = lower_position_violations.nonzero()[0][0]
-                raise JointLimitError(robot,
-                    dof_index=active_indices[index],
-                    dof_value=q[index],
-                    dof_limit=q_limit_min[index],
-                    description='position')
-
-            upper_position_violations = (q > q_limit_max)
-            if upper_position_violations.any():
-                index = upper_position_violations.nonzero()[0][0]
-                raise JointLimitError(robot,
-                    dof_index=active_indices[index],
-                    dof_value=q[index],
-                    dof_limit=q_limit_max[index],
-                    description='position')
+            # Check joint position limits.
+            # We do this before setting the joint angles.
+            util.CheckJointLimits(robot, q)
 
             robot.SetActiveDOFValues(q)
 
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -113,7 +113,8 @@ class VectorFieldPlanner(BasePlanner):
 
             Succeed if end-effector pose is close to the goal pose.
             """
-            pose_error = util.GeodesicDistance(manip.GetEndEffectorTransform(),
+            pose_error = util.GetGeodesicDistanceBetweenTransforms(
+                                               manip.GetEndEffectorTransform(),
                                                goal_pose)
             if pose_error < pose_error_tol:
                 return Status.TERMINATE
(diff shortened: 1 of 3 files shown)


DavidB-CMU:
#### David wrote:  
General questions about VectorFieldPlanner:  
- If you call PlanToEndEffectorOffset() for some distance, the planner never reaches that distance.  Can we fix this?  
- Should we rename the "max_distance" argument? it seems confusing to me.  
- Potentially PlanToEndEffectorOffset() could be refactored to call PlanWorkspacePath, however I think PlanToEndEffectorOffset() has a more efficient implementation.  
- Do all the PlanWorkspacePath() methods in each planner need to have the same prototype?  
- To implement PlanWorkspacePath(), I call GetMinDistance... in both of the callbacks, so maybe this is not optimal.

#### Mike replied:  
This is a very nice addition. I left a bunch of comments on the diff and have answered your questions below:

> If you call PlanToEndEffectorOffset() for some distance, the planner never reaches that distance. Can we fix this?

Can you be more specific?

> Should we rename the "max_distance" argument? it seems confusing to me.

What do you suggest as a new name? I am open to renaming this as long as:

1. we add a stub that maintains backwards compatibility; i.e. forward the `max_distance` parameter and raise a deprecation warning
2. update any other `PlanToEndEffectorOffset` methods that take this argument in the same way

> Potentially PlanToEndEffectorOffset() could be refactored to call PlanWorkspacePath, however I think PlanToEndEffectorOffset() has a more efficient implementation.
>
> To implement PlanWorkspacePath(), I call GetMinDistance... in both of the callbacks, so maybe this is not optimal.

That would be a nice change as long as, like you said, it does not have major performance implications. Can you time the two implementations and see if there's a noticeable performance difference?

> Do all the PlanWorkspacePath() methods in each planner need to have the same prototype?

The same arguments get passed to all planners in a `Sequence`. At a minimum, all of the planners should have identical *required arguments*. It's fine to have different optional arguments, but we do try to make them overlap as much as possible (e.g. we consistently use `timelimit` for all planners that have a planning time limit).

#### David wrote: 
I will focus on finishing this PR (see below), then we can come back to the above comments.

DavidB-CMU:
#### Mike's diff comments:  

**No. 1**:  

```
def vf_geodesic():
"""
Function defining a joint-space vector field.
"""
```

This __docstring__ is misleading. An arbitrary vector field function implements a joint-space vector field. This function specifically implements a vector field that moves along the geodesic towards goal_pose.

I also don't think this function strictly needs a __doscstring__ - although I'm certainly not going to argue with adding one - since it is only accessible inside this function.

=========

**No. 2**:  

```
def vf_geodesic():
return min(abs(vlimits[i] / dqout[i]) if dqout[i] != 0. else 1. for i in xrange(vlimits.shape[0])) * dqout
 
         def CloseEnough():
-            pose_error = util.GeodesicDistance(
-                        manip.GetEndEffectorTransform(),
-                        goal_pose)
+            """
+            Function defining the termination condition.
+
+            Succeed if end-effector pose is close to the goal pose.
+            """

```

See my comment above on vf_geodesic. The same applies here.

=======

**No. 3**:  

```
return Status.CONTINUE
 
+        integration_timelimit = 10.0
```

I believe this should use the timelimit argument.

=============

**No. 4**:  

```
+        # Get the final end-effector pose
+        duration = traj.GetDuration()
+        T_ee_goal = openravepy.matrixFromPose(traj.Sample(duration)[0:7])
+
+        def vf_path():
+            """
+            Function defining a joint-space vector field.
+            """
+            T_ee_actual = manip.GetEndEffectorTransform()
+
+            # Find where we are on the goal trajectory by finding
+            # the the closest point
+            (_,t) = util.GetMinDistanceBetweenTransformAndWorkspaceTraj(
+                                                                   T_ee_actual,
+                                                                   traj)

```

I believe this will fail if the trajectory has a loop (or near-loop) in it. You may want to mention this in the __docstring__.

======

**No. 5**:  

```
+        @param float angular_tolerance:  Constraint tolerance (radians).
+        @param float goal_pose_tol: Goal tolerance (meters).
+
+        @return openravepy.Trajectory qtraj: Configuration space path.
+        """
+        if position_tolerance < 0.0:
+            raise ValueError('Position tolerance must be non-negative.')
+        elif angular_tolerance < 0.0:
+            raise ValueError('Angular tolerance must be non-negative.')
+
+        manip = robot.GetActiveManipulator()
+        Tstart = manip.GetEndEffectorTransform()
+
+        # Get the final end-effector pose
+        duration = traj.GetDuration()
+        T_ee_goal = openravepy.matrixFromPose(traj.Sample(duration)[0:7])
```

This assumes that the trajectory has a single AffineDOFs group that is parameterized by a translation and a quaternion. If you are making this assumption, then you should check for this when you validate the input arguments (and add a few unit tests).

Also, if this is a path (i.e. untimed), how are you Sampleing with a time?

===

**No. 6**:  

```
+        def vf_path():
+            """
+            Function defining a joint-space vector field.
+            """
+            T_ee_actual = manip.GetEndEffectorTransform()
+
+            # Find where we are on the goal trajectory by finding
+            # the the closest point
+            (_,t) = util.GetMinDistanceBetweenTransformAndWorkspaceTraj(
+                                                                   T_ee_actual,
+                                                                   traj)
+            # Get the desired end-effector transform from
+            # the goal trajectory
+            t_step = 0.2
+            desired_T_ee = openravepy.matrixFromPose(traj.Sample(t)[0:7])
+            pose_ee_next = traj.Sample(t+t_step)[0:7]
```

It looks like you are using t_step to approximate a finite-difference velocity. This is problematic because:  
- the trajectory is un-timed (see above)
- you have no guarantee on the scale of the parameterization - 0.2 may be very large or very small

Ideally, we would directly extract the derivative from the trajectory and use this as the feed-forward term in the controller. If that is not possible, then we should at least convert the trajectory to a known parametrization (e.g. unit velocity) so we know the scale of t_step

======

**No. 7**:  

```
+            # on the goal trajectory
+            twist_parallel = util.GeodesicTwist(desired_T_ee,
+                                                desired_T_ee_next)
+            direction_parallel = desired_T_ee_next[0:3,3] - desired_T_ee[0:3,3]
+
+            twist = 0.0*twist_perpendicular + 1.0*twist_parallel
+
+            K_p = 0.5
+            direction = (1.0-K_p)*direction_perpendicular + \
+                         K_p*direction_parallel
+            # Normalize the direction vector
+            direction = numpy.array(direction, dtype='float')
+            direction /= numpy.linalg.norm(direction)
+
+            # Modify the twist
+            twist[0:3] = direction
```

I'm not sure whats going on here. I think the current implementation will allow arbitrary drift of orientation, since the (1 - Kp) * direction_parallel feedback term is only applied to the translational component.

======

**No. 8**:  

```
+            twist = 0.0*twist_perpendicular + 1.0*twist_parallel
+
+            K_p = 0.5
+            direction = (1.0-K_p)*direction_perpendicular + \
+                         K_p*direction_parallel
+            # Normalize the direction vector
+            direction = numpy.array(direction, dtype='float')
+            direction /= numpy.linalg.norm(direction)
+
+            # Modify the twist
+            twist[0:3] = direction
+
+            # Calculate joint velocities using an optimized jacobian
+            dqout, _ = util.ComputeJointVelocityFromTwist(robot, twist,
+                                              joint_velocity_limits=numpy.PINF)
```

Why are you passing infinite velocity limits here? This is concerning for a few reasons:  
- Isn't joint_velocity_limits typically a vector of length len(robot.GetActiveDOFs()), not a scalar?
- Does this function behave reasonably when the value is set to infinity?
- Won't this result in very large steps?

========

**No. 9**:  

```
+                    'Deviated from orientation constraint.')
+
+            position_deviation = numpy.linalg.norm(position_error)
+            if position_deviation > position_tolerance:
+                raise ConstraintViolationPlanningError(
+                    'Deviated from straight line constraint.')
+
+            # Check if we have reached the end of the goal trajectory
+            dist_to_goal = util.GetGeodesicDistanceBetweenTransforms(T_ee_curr,
+                                                                     T_ee_goal)
+            if numpy.fabs(dist_to_goal) <= goal_pose_tol:
+                return Status.CACHE_AND_TERMINATE
+
+            return Status.CONTINUE
+
+        integration_timelimit = 10.0
```

This should be an argument.

========

**No. 10**:  

```
return fn_vectorfield()
 
         def fn_status_callback(t, q):
+            """
+            This is called for each collision check.
+            """
```

This is overly specific. More than just collision checking happens in here - e.g. constraint violation checks.

=====

**No. 11**:  

```
"""
         Follow a joint space vectorfield to termination.
 
         @param robot
         @param fn_vectorfield a vectorfield of joint velocities
         @param fn_terminate custom termination condition
+        @param integration_timelimit The initial time step to be taken
+                                     by the integrator.
```

This __docstring__ is misleading. The initial step size is passed as the first_step parameter to the integrator and is currently hard-coded to 0.1. This argument appears to be the value of t for which we're trying to compute the value of the integral.

===

**No. 12**:  

```
+
+def GetGeodesicDistanceBetweenTransforms(T0, T1):
+    """
+    Wrapper, to match GetGeodesicDistanceBetweenQuaternions()
+    """
+    return GeodesicDistance(T0, T1, r=1.0)
+
+
+def GetEuclideanDistanceBetweenTransforms(T0, T1):
+    """
+    Calculate the Euclidean distance between two 4x4 transforms.
+    (also called L2 or Pythagorean distance)
```

This appears to only consider translational distance. This is not clear from the name of the function of the __docstring__.

=====

**No. 13**:  

```
+    if not IsWorkspaceTrajectory(traj):
+        raise ValueError("Trajectory is not a workspace trajectory, it "
+                         "must have configuration specification of "
+                         "openravepy.IkParameterizationType.Transform6D")
+
+    min_dist = numpy.inf
+    t_loc = 0.0
+    t = 0.0
+    while t < traj.GetDuration():
+        T_curr = openravepy.matrixFromPose(traj.Sample(t)[0:7])
+        error = GetEuclideanDistanceBetweenTransforms(T, T_curr)
+        if error < min_dist:
+            min_dist = error
+            t_loc = t
+        t = t + dt
+    return (min_dist, t_loc)
```

You may want to add a special case that explicitly checks traj.GetDuration(). Otherwise, I could imagine this messing up the termination condition above.

It would also be useful to return T_curr in the tuple so the user doesn't have to evaluate the trajectory again.

=====

DavidB-CMU:
Most of the above items have been addressed.  
Still outstanding:  
 - Implement new function: util.ComputeGeodesicUnitTiming()  
 - Another new function: util.GetAngularVelocity() for trajectory of type IkParametrization  
 - add unit tests: pass in a non-workspace trajectory, etc.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +219,197 @@ def TerminateMove():
...
+        """
+
+        if not util.IsTrajectoryTypeIkParameterization(traj):
+            raise ValueError("Trajectory is not a workspace trajectory, it "
+                             "must have configuration specification of "
+                             "openravepy.IkParameterizationType.Transform6D")
+
+        #if IsTimedTrajectory(traj):
+        #    raise ValueError("PlanWorkspacePath expected an un-timed "
+        #                     "trajectory.")
```
mkoval: Please remove this comment.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +219,197 @@ def TerminateMove():
...
+        elif angular_tolerance < 0.0:
+            raise ValueError('Angular tolerance must be non-negative.')
+        elif goal_pose_tol < 0.0:
+            raise ValueError('Goal-pose tolerance must be non-negative.')
+
+        #
+        #
+        # ADD ComputeGeodesicUnitTiming()
+        #
+        #
```
mkoval: I believe this is necessary for the function to work, since you no longer require that the input `path` is timed.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +219,197 @@ def TerminateMove():
...
+            T_ee_actual = manip.GetEndEffectorTransform()
+
+            # Find where we are on the goal trajectory by finding
+            # the the closest point
+            (_,t,_) = util.GetMinDistanceBetweenTransformAndWorkspaceTraj(
+                                                                   T_ee_actual,
+                                                                   traj)
+            # Get the desired end-effector transform from
+            # the goal trajectory
+            t_step = 0.2
```
mkoval: This constant still makes me queasy. Could you at least make it an argument?

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +219,197 @@ def TerminateMove():
...
+                    'Deviated from orientation constraint.')
+
+            position_deviation = numpy.linalg.norm(position_error)
+            if position_deviation > position_tolerance:
+                raise ConstraintViolationPlanningError(
+                    'Deviated from straight line constraint.')
+
+            # Check if we have reached the end of the goal trajectory
+            dist_to_goal = util.GetGeodesicDistanceBetweenTransforms(T_ee_curr,
+                                                                     T_ee_goal)
```
mkoval: It's strange that constraint deviation is specified separately for position and orientation, but the goal tolerance is specified as a single distance threshold. Both checks should be consistent. I slightly prefer the independent checks, since having separate error messages makes it easier to diagnose why planning failed.

src/prpy/planning/vectorfield.py
```diff
@@ -197,32 +219,197 @@ def TerminateMove():
...
+            if numpy.fabs(dist_to_goal) <= goal_pose_tol:
+                return Status.CACHE_AND_TERMINATE
+
+            return Status.CONTINUE
+
+        integration_interval = 10.0
```
mkoval: Please make this an argument.

src/prpy/planning/vectorfield.py
```diff
@@ -240,45 +427,46 @@ def FollowVectorField(self, robot, fn_vectorfield, fn_terminate,
...
             return fn_vectorfield()
 
         def fn_status_callback(t, q):
+            """
+            Check joint-limits and collisions for a specific joint
+            configuration. This is called multiple times at DOF
+            resolution in order to check along the entire length of the
+            trajectory.
+            Note: Currently this is called after each integration time
+            step, which means we are doing more checks than required.
```
mkoval: This comment is not true. The `fn_callback` function below is called after every integration timestep, but only calls `fn_status_callback` if the robot has moved more than DOF resolution. If that is not the case, then it is a bug - possibly because of the comment about `GetCollisionCheckPts` below.

src/prpy/util.py
```diff
@@ -196,6 +196,19 @@ def ComputeAinv(N,dof):
                 invA[i*dof+k,j*dof+k] = invA_small[i-1,j-1]
     return invA
     
+def NormalizeVector(vec):
+    """
+    Normalize a vector.
+    This is faster than doing: vec/numpy.linalg.norm(vec)
```
mkoval: Is this really a bottleneck? I'd be quite surprised.

src/prpy/util.py
```diff
@@ -725,6 +738,75 @@ def GeodesicDistance(t1, t2, r=1.0):
     error[3] = r*error[3]
     return numpy.linalg.norm(error)
 
+
+def GetGeodesicDistanceBetweenTransforms(T0, T1):
+    """
+    Wrapper, to match GetGeodesicDistanceBetweenQuaternions()
+    """
+
+    return GeodesicDistance(T0, T1, r=1.0)
```
mkoval: I'm not thrilled about hard-coding `r=1.0` here. This parameter has a huge effect on the behavior of the distance function.

src/prpy/util.py
```diff
@@ -725,6 +738,75 @@ def GeodesicDistance(t1, t2, r=1.0):
...
+
+def GetEuclideanDistanceBetweenTransforms(T0, T1):
+    """
+    Calculate the Euclidean distance between the translational
+    component of two 4x4 transforms.
+    (also called L2 or Pythagorean distance)
+    """
+    p0 = T0[0:3,3] # Get the x,y,z translation from the 4x4 matrix
+    p1 = T1[0:3,3]
+    return numpy.sqrt(numpy.sum((p0-p1)**2))
```
mkoval: Why not use `numpy.linalg.norm` here?

src/prpy/util.py
```diff
@@ -897,6 +979,37 @@ def IsTimedTrajectory(trajectory):
...
+
+def IsTrajectoryTypeIkParameterization(traj):
+    """
+    Check if trajectory is a workspace trajectory.
+
+    @param openravepy.Trajectory traj: A path or trajectory.
+    @return bool result: Returns True or False.
+    """
+    try:
+        if traj.GetConfigurationSpecification().GetGroupFromName("ikparam_values"):
```
mkoval: You should also check the *type* of the `IkParameterization` here. I believe this is as simple as appending the name of the parameterization to the string you pass to `GetGroupFromName`.

mkoval:
I think this is very close to being ready to merge. I added a few comments about minor quibbles above, but the only major issue remaining is to add unit tests and the `ComputeGeodesicUnitTiming` function.

I think it's okay to keep the finite-differenced velocities for now, as long as you can guarantee that the trajectory has unit timing. Adding support for derivatives is likely more trouble than it's worth - we don't even have a good way to generating a timing for an `IkParameterization` trajectory in the first place.

DavidB-CMU:
All of the above have been addressed.
I'm working on the Unit Tests.
@mkoval Can you have a look at this please?

DavidB-CMU:
Added Unit Tests for new functions in prpy.util

mkoval:
:+1: Thanks for the great unit tests as always!

PR #270, opened 2016-02-22
Added missing SetActiveDOF in VectorField.

psigen:
This fixes an ongoing issue with VectorFieldPlanner immediately failing
in certain cases (which turned out to be cases where the Active DOFs
set on the robot did not match the Active Manipulator).

I also removed some unused imports.

### DIFF
--- a/src/prpy/planning/retimer.py
+++ b/src/prpy/planning/retimer.py
@@ -95,8 +95,8 @@ class OpenRAVERetimer(BasePlanner):
 
         # Remove co-linear waypoints. Some of the default OpenRAVE retimers do
         # not perform this check internally (e.g. ParabolicTrajectoryRetimer).
-        if not IsTimedTrajectory(output_traj):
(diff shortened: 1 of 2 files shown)


mkoval:
This is not a bug. Active DOFs and the active manipulator are independent:

1. active DOFs represent the DOFs that may be modified by the planner
2. the active manipulator defines the end-effector frame that is used in the objective function.

It's perfectly reasonable to set the active DOFs to a subset of the active manipulator's `GetArmIndices()` if you only want to plan for those DOFs. It's also valid, although not useful, to set the active DOFs to joints that do not affect the end effector.

If the planner does not work when these are not equal, which is very possible, then *that is a bug*. We *should not* work around the bug by setting the active DOFs to those of the active manipulator.

mkoval:
@psigen What was our conclusion on this?

psigen:
I need to refactor this to do a check instead of setting the DOFs.

psigen:
This PR is being used for reference, but should not be merged.

PR #271, opened 2016-02-25
Adding ability to get a list of named configurations

jeking04:
This pull request allows you to get a list of known configurations via:
```
named_configs = robot.configurations.get_configuration_list()
```

### DIFF
--- a/src/prpy/named_config.py
+++ b/src/prpy/named_config.py
@@ -112,3 +112,9 @@ the following format:
             return self._configs[name]
         else:
             raise KeyError('There is no configuration named %s.' % name)
+    
+    """
+    : return known_configurations: A list of named configurations
+    """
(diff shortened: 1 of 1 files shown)


psigen:
It seems to me that 'configurations' and 'tsrs' sound like they are list-like or dict-like, and it is really counter-intuitive for them to have completely custom APIs.  Having a call like `robot.configurations.get_configuration_list()` really seems awkward and not what I would expect to have to do.

But I don't really use this API much, so I'll leave those decisions to you guys.

mkoval:
@psigen I posted a justification for the `ConfigurationLibrary` class in https://github.com/personalrobotics/herbpy/pull/81. As always, I'm welcome to hear a compelling counter-argument. :smile:

Issue #256, opened 2016-01-14
Print warning if (attempting) to double clone an environment / Refactor workspace.py


DavidB-CMU:
This is related to pull request https://github.com/personalrobotics/prpy/pull/55 ("Fixed a number of bugs related to workspace planner") 

If you call a planning method like PlanToEndEffectorOffset() it then calls PlanWorkspacePath(). 
This results in two calls to Clone(), but there's a check in there
```python
if self.clone_env != self.clone_parent:
    with self.clone_parent:
        self.clone_env.Clone(self.clone_parent, self.options)
```
which means that the environment is not cloned twice.

In https://github.com/personalrobotics/prpy/pull/55 @mkoval suggested that we should print a warning message like "Warning: you are trying to clone an environment into itself"
Should we add this warning message? 

I would like to suggest that we also re-factor `workspace.py` so Clone() is not called twice.

psigen:
After doing some work on this in #279, I think the best option here is to simply make cloning into yourself be a near-free operation.

There are a lot of cases, like nested `@PlanningMethod` calls, where deduplicating cloning is a very verbose operation, either requiring refactoring tons of common code into private helper methods, or branching off a special case when the clone is not required.

Since `with Clone()` also performs locking and adds a `.Cloned()` operator which normalized references which may or may not be in the environment, it already handles a lot of edge cases for us when we use it.  I think cloning into the same environment should be one of those cases, so we should make it so that there is no penalty for doing so.

Semantically, we will just be accessing/locking the one environment we would be using anyway, so I think this is a safe thing to do.

mkoval:
:+1: I agree with @psigen analysis. Eliminating all nested `@PlanningMethod` introduces too much boilerplate to be a viable solution.

I suspect that #279 eliminates most overhead. I am going to close this issue, but please do not hesitate to re-open it if you notice performance issues from multiple `Clone` calls into the same environment.

PR #272, opened 2016-03-07
Removed duplicate @PlanningMethod decorators.

psigen:
This commit removes several instances of nested @PlanningMethod
decorators on several of our planners.  It refactors some methods
into internal calls that are wrapped by a new @PlanningMethod instead
so they can be called from multiple @PlanningMethods without duplication.

### DIFF
--- a/src/prpy/planning/retimer.py
+++ b/src/prpy/planning/retimer.py
@@ -154,7 +154,6 @@ class HauserParabolicSmoother(OpenRAVERetimer):
             'time_limit': float(timelimit),
         })
 
-    @PlanningMethod
     def RetimeTrajectory(self, robot, path, options=None, **kw_args):
         from copy import deepcopy
         if options is None:
@@ -163,6 +162,7 @@ class HauserParabolicSmoother(OpenRAVERetimer):
         new_options['time_limit'] = kw_args.get('timelimit', 3.)
         return super(HauserParabolicSmoother, self).RetimeTrajectory(
             robot, path, options=new_options, **kw_args)
+
 
 class OpenRAVEAffineRetimer(BasePlanner):
     def __init__(self,):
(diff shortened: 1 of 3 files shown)


src/prpy/planning/retimer.py
```diff
@@ -154,7 +154,6 @@ def __init__(self, do_blend=True, do_shortcut=True, blend_radius=0.5,
             'time_limit': float(timelimit),
         })
 
-    @PlanningMethod
     def RetimeTrajectory(self, robot, path, options=None, **kw_args):
```
mkoval: Won't this break using `HauserParabolicSmoother` in a `Sequence`?

mkoval:
:+1: Great catch. I have only have one concern about `HauserParabolicSmoother` before I merge this (see above).

It looks like Travis failed due to a transient error. I restarted the build.

DavidB-CMU:
Thanks @pras, this will address part of issue https://github.com/personalrobotics/prpy/issues/256 I think.

psigen:
Oops, yeah that would have messed up `SequencePlanner`.  I refactored it to use an internal method like the others.

mkoval:
LGTM. I'm happy to merge after you resolve the conflict. It's likely just that `_FollowVectorField` rename is conflicting with the recent merge of #269.

psigen:
@mkoval can you restart the checks? I think it failed for no reason.
https://travis-ci.org/personalrobotics/prpy/builds/114531617

psigen:
@mkoval: Actually, is it better to modify @PlanningMethod to be re-entrant?
In theory we can avoid re-cloning if we do a few checks in key places.

mkoval:
That would be a better solution. It's not immediately obvious to me how to do it, but go ahead if you have an idea!

mkoval:
I restarted the build again. I'm not sure why Travis has been so unhappy.

PR #274, opened 2016-03-16
Remove HauserParabolicSmoother as default smoother.

psigen:
This removes HauserParabolicSmoother as a default smoother inside `prpy.base.robot`. 

Having `HauserParabolicSmoother` hard-coded in this `__init__()` means that _any_ `prpy` robot depends on the external package `or_parabolicsmoother`, even if they do not use it at all in their subclass implementation.

Since we can easily override this default in subclassed (and almost always do to customize the retiming settings anyway), there is no reason to force this dependency.


mkoval:
I just verified that our `*py` implementations do indeed override the defaults. LGTM.

PR #276, opened 2016-03-21
Added PEP8 fixes and lint corrections for `prpy.planning`.

psigen:
Fixes a number of small syntax errors and non-compliant formatting in the `prpy.planning` modules.


mkoval:
Thanks for putting this in its own pull request!

PR #277, opened 2016-03-24
Fixed HauserParabolicSmoother overriding timelimit

psigen:
This fixes a bug where the HauserParabolicSmoother was ignoring `timelimit` settings in the default options in the `RetimeTrajectory` call.

### DIFF
--- a/src/prpy/planning/retimer.py
+++ b/src/prpy/planning/retimer.py
@@ -30,6 +30,7 @@
 
 import logging
 import openravepy
+from copy import deepcopy
 from ..util import (CreatePlannerParametersString, CopyTrajectory,
                     SimplifyTrajectory, HasAffineDOFs, IsTimedTrajectory)
 from base import (BasePlanner, PlanningError, PlanningMethod,
@@ -156,11 +157,9 @@ class HauserParabolicSmoother(OpenRAVERetimer):
 
     @PlanningMethod
     def RetimeTrajectory(self, robot, path, options=None, **kw_args):
-        from copy import deepcopy
-        if options is None:
-            options = {}
-        new_options = deepcopy(options)
-        new_options['time_limit'] = kw_args.get('timelimit', 3.)
+        new_options = deepcopy(options) if options else dict()
+        if 'timelimit' in kw_args:
+          new_options['time_limit'] = kw_args['timelimit']
         return super(HauserParabolicSmoother, self).RetimeTrajectory(
             robot, path, options=new_options, **kw_args)
 


src/prpy/planning/retimer.py
```diff
@@ -156,11 +157,9 @@ def __init__(self, do_blend=True, do_shortcut=True, blend_radius=0.5,
 
     @PlanningMethod
     def RetimeTrajectory(self, robot, path, options=None, **kw_args):
-        from copy import deepcopy
-        if options is None:
-            options = {}
-        new_options = deepcopy(options)
-        new_options['time_limit'] = kw_args.get('timelimit', 3.)
+        new_options = deepcopy(options) if options else dict()
```
mkoval: I don't understand what the point of the `options` argument is at all. Why aren't we simply using `kw_args` for this?

@psigen, I know that you didn't add this in the pull request, but do you understand what the logic is here?

jeking04:
@mkoval For what it is worth, I am pretty sure we added that options parameter when originally fixing this ```time_limit```/```timelimit``` issue. But I can't remember why we thought adding a new dictionary was better than just using ```kw_args```. One thought is that maybe the smoother doesn't gracefully handle having unrecognized parameters (i.e. parameters other than the ones we specified in ```default_options```). The ```options``` dictionary doesn't specifically protect against that, but it makes it far less likely than just dumping everything in ```kw_args``` into the parameters.  But again, this is just a guess rooted in a vague recollection of our original fix.  Seems like we could just use ```kw_args```.

psigen:
Pretty sure @jeking04 is on the right track here.  IIRC, the issue is here:
https://github.com/personalrobotics/prpy/blob/master/src/prpy/planning/retimer.py#L89

We blindly create the XML parameters for the planner using _every_ key in the options structure.  However, OpenRAVE planners will completely reject XML parameters they don't handle.  At least, this seemed to be the case when I misspelled 'blend_radius' in my options.

Since we had no way to distinguish things that should go to the retimer's PlannerParameters from things that go to `RetimeTrajectory` (which might include meta-information for other operations, like the way the `execute` flag works for `Plan` calls), we decided to put the `options` dict() in.  If we can figure out a way to make OpenRAVE not reject unmatched XML tags, then we could remove `options`.

mkoval:
:+1: This change looks good, then.

cdellin:
FWIW, OpenRAVE planners generally ignore XML parameters that they don't support (including built-in planners, CBiRRT, LEMUR, etc).  This is usually implemented via the `_vXMLParameters` member [0].  The one exception to this is `or_ompl`, which doesn't use the OpenRAVE XML parsing code.  Instead, it parses the XML itself, passes each parameter to the underlying OMPL planner, and errors if the planner doesn't support a particular option.  It might be worthwhile the fix the way it handles unsupported options.

[0] http://openrave.org/docs/latest_stable/coreapihtml/classOpenRAVE_1_1PlannerBase_1_1PlannerParameters.html#ad56d63f8996ddbeda0f31326154e9e94

psigen:
I'm happy to change this over in a followup PR, if everyone is OK with it.

mkoval:
I am strongly *against* silently ignoring unsupported parameters. It's *far* too easy to misspell an argument name (e.g. `time_limit` or `timelimit`?) and have it silently ignored. This is difficult to to debug if you don't notice that it's happening.

cdellin:
I definitely see @mkoval's point here, I was just talking about what I think the API intends.  Perhaps a middle-ground would be to print a warning, or to handle unknown parameters at the Python level?

psigen:
Actually, I don't think having `options` actually solves this problem at all, we've just never encountered any situation with multiple smoothers.

Say I have `HauserParabolicSmoother` and `KunzCircularBlender` in a `Sequence()` to do `RetimeTrajectory`.  `KunzCircularBlender` takes a `max_deviation`, and `HauserParabolicSmoother` takes a `blend_radius`.

How do I construct an `options` dictionary that 
1. will detect unused arguments 
2. will not simply spam me warnings about the respective planners ignoring each others options
3. does not require creating a unique options dictionary for every planner that might be in the Sequence?

I think this is the reason we switched to using `kwargs` in the planning stack in the first place.

PR #278, opened 2016-03-25
Removes the ill-fated Trollius support from PlanningMethod.

psigen:
Trollius support never actually worked very well (or at all) for our planners, and with the addition of `prpy.futures` we have a much better interface for doing parallel operations.

This removes the `defer` options on all prpy planning methods, and replaced the `Ranked` implementation with one that uses `prpy.futures` and plain threading instead.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -370,7 +370,6 @@ class Robot(openravepy.Robot):
 
         # Copy the trajectory into the output environment.
         output_traj = CopyTrajectory(traj, env=self.GetEnv())
-        output_traj.SetDescription(path.GetDescription())
         return output_traj
 
     def ExecutePath(self, path, **kwargs):
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -6,7 +6,7 @@
 #
 # Redistribution and use in source and binary forms, with or without
 # modification, are permitted provided that the following conditions are met:
-# 
+#
 # - Redistributions of source code must retain the above copyright notice, this
 #   list of conditions and the following disclaimer.
 # - Redistributions in binary form must reproduce the above copyright notice,
@@ -15,7 +15,7 @@
 # - Neither the name of Carnegie Mellon University nor the names of its
 #   contributors may be used to endorse or promote products derived from this
 #   software without specific prior written permission.
-# 
+#
 # THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
 # AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
 # IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
@@ -211,7 +211,7 @@ class Robot(openravepy.Robot):
 
         return active_manipulators
 
-    def PostProcessPath(self, path, defer=False, executor=None,
+    def PostProcessPath(self, path,
                         constrained=None, smooth=None, default_timelimit=0.5,
                         shortcut_options=None, smoothing_options=None,
                         retiming_options=None, affine_retiming_options=None,
@@ -237,7 +237,7 @@ class Robot(openravepy.Robot):
         The behavior in (2) and (3) can be forced by passing constrained=True
         or smooth=True. By default, the case is inferred by the tag(s) attached
         to the trajectory: (1) is triggered by the CONSTRAINED tag and (2) is
-        tiggered by the SMOOTH tag.
+        triggered by the SMOOTH tag.
 
         Options an be passed to each post-processing routine using the
         affine_retiming_options, shortcut_options, smoothing_options, and
@@ -245,8 +245,6 @@ class Robot(openravepy.Robot):
         in any of these dictionaries, it defaults to default_timelimit seconds.
 
         @param path un-timed OpenRAVE trajectory
-        @param defer return immediately with a future trajectory
-        @param executor executor to use when defer = True
         @param constrained the path is constrained; do not change it
         @param smooth the path is smooth; attempt to execute it directly
         @param default_timelimit timelimit for all operations, if not set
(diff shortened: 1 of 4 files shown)


src/prpy/base/robot.py
```diff
@@ -288,107 +286,94 @@ def PostProcessPath(self, path, defer=False, executor=None,
...
+            logger.debug(
+                'Setting robot "%s" DOFs %s (affine? %d) as active for'
+                ' post-processing.',
+                cloned_robot.GetName(), list(dof_indices), affine_dofs
+            )
+
+            if len(dof_indices) and affine_dofs:
+                raise ValueError(
+                    'Trajectory contains both affine and regular DOFs.')
+            # Special case for timing affine-only trajectories.
```
mkoval: Missing newline before this comment.

src/prpy/base/robot.py
```diff
@@ -288,107 +286,94 @@ def PostProcessPath(self, path, defer=False, executor=None,
...
-            return do_postprocess()
-        else:
-            raise ValueError('Received unexpected value "{:s}" for defer.'.format(defer))
+                    logger.debug('Smoothing an unconstrained path.')
+                    traj = self.smoother.RetimeTrajectory(
+                        cloned_robot, shortcut_path, **smoothing_options)
 
+        # Copy the trajectory into the output environment.
+        output_traj = CopyTrajectory(traj, env=self.GetEnv())
+        output_traj.SetDescription(path.GetDescription())
```
mkoval: Does `CopyTrajectory` *not* copy `GetDescription`? That sounds like a bug.

src/prpy/planning/base.py
```diff
@@ -332,62 +309,44 @@ def get_planners(self, method_name):
...
             else:
-                planners.append((index, planner))
+                futures[index] = defer(call_planner, args=(planner,))
 
-        # Helper function to call a planner and store its result or error.
-        def call_planner(index, planner):
+        # Each time a planner completes, check if we have a valid result
+        # (a planner found a solution and all higher-ranked planners had
+        # already failed).
+        for index, future in futures.viewitems():
```
mkoval: Python `dict`s are not ordered by key, so this *does not* call `result()` on the futures in the correct order. I suggest replacing `futures` with a `list` of `tuples` since you never query by index.

mkoval: Also, I am not familiar with `viewitems`. Is there an advantage over `iteritems` in this context?

src/prpy/base/robot.py
```diff
@@ -288,107 +286,94 @@ def PostProcessPath(self, path, defer=False, executor=None,
...
 
-                cloned_robot.SetActiveDOFs(dof_indices, affine_dofs)
-                logger.debug(
-                    'Setting robot "%s" DOFs %s (affine? %d) as active for'
-                    ' post-processing.',
-                    cloned_robot.GetName(), list(dof_indices), affine_dofs
-                )
+        with Clone(self.GetEnv(),
+                   clone_env=self._postprocess_env) as cloned_env:
+            cloned_robot = cloned_env.Cloned(self)
```
mkoval: Cloning isn't technically necessary at all now that we've removed `defer` support. Did you add this extra clone for a particular reason?

mkoval:
This is a fantastic change - our ill-fated attempt at using Trollius introduced a lot of needless complexity into the planning pipeline. I am grateful that you removed it and left some specific comments above.

@jeking04 was attempting to use `defer=True` earlier this week. I'd like to check with her before merging this pull request to see which features, if any, she is currently using.

jeking04:
I am good with merging this. I couldn't ever quite figure it out and ended
up just using threading around the execute trajectory call rather than the
defer flag.

On Saturday, March 26, 2016, Michael Koval <notifications@github.com> wrote:

> This is a fantastic change - our ill-fated attempt at using Trollius
> introduced a lot of needless complexity into the planning pipeline. I am
> grateful that you removed it and left some specific comments above.
>
> @jeking04 <https://github.com/jeking04> was attempting to use defer=True
> earlier this week. I'd like to check with her before merging this pull
> request to see which features, if any, she is currently using.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/personalrobotics/prpy/pull/278#issuecomment-201707494>
>


-- 
Jen King

src/prpy/planning/base.py
psigen: `Views` are lazily-constructed iterators and are the default in Python 3:
http://legacy.python.org/dev/peps/pep-0469/#set-based-dynamic-views
http://stackoverflow.com/a/23083560

In theory, they have a slight performance advantage, but I've mostly started using them because they share semantics with the clearly favored `.items()` `.keys()` and `.values()` methods in Python 3.

src/prpy/base/robot.py
psigen: I didn't add it, it was just moved from the top of `do_postprocess()`.  Why does removing `defer` remove the need to clone?

src/prpy/base/robot.py
psigen: I don't understand, since I just re-indented this code without changing it.

Where do you want a newline?  Do you want a blank line before an `elif`?

src/prpy/base/robot.py
psigen: Addressed separately in #282, so I will remove this here.

src/prpy/planning/base.py
psigen: Addressed in d9a9d40.

src/prpy/base/robot.py
psigen: In fact, I don't see how we could do this otherwise, we need to set active DOFs on the robot before we call the retimers.  How do we do that without either locking the `env` for the duration of the `PostProcessPath()` call or using a clone?

mkoval: We needed the clone when `defer=True` because post-processing may be a two stage process. We first call the shortcutter, then call the retimer (or smoother). Cloning is necessary to guarantee that the environment is the same between the two calls.

I see your point. Nothing has fundamentally changed about this, so we're stuck with the clone.

src/prpy/base/robot.py
mkoval: It looks a bit odd to have no newline between the `ValueError` and this comment. I suppose it's fine, though, because we do this all the time on `elif`s.

mkoval:
Thanks for the input, @jeking04! It sounds like you agree that this was an ill-fated endeavor. :sweat:

I am good with merging this after #282.

PR #279, opened 2016-03-25
Moved regrab logic into env equality check.

psigen:
Previously, we did a relatively expensive self-collision and regrab checks on the robot in *any* `Clone()` operation, on the off-chance that it was grabbing something that could get messed up.

This moves the `CheckSelfCollision()` and `RegrabAll()` operations inside the following checks:

1. We are not cloning back into the same environment.
2. The robot is actually grabbing something.

This shaves time off nested cloning calls, which are found everywhere in our planning stack.  It may also obviate the need for #272.


mkoval:
:+1: Excellent change.

PR #280, opened 2016-03-25
Added separate LockedPlanningMethod and ClonedPlanningMethod calls.

psigen:
**This should be merged _after_ #278 and #279.**

This separates the `@PlanningMethod` decorator into two decorators:

- `@LockedPlanningMethod` locks the current robot environment and executes the wrapped method.
- `@ClonedPlanningMethod` clones the current robot environment and executes the wrapped method.

This distinction is useful for planning calls that are so fast that they take less time than the cloning operation itself, such as optimistic plans like `SnapPlanner` and fast mathematical solutions such as `KunzCircularBlender`.

This commit refactors all existing planning calls except `SnapPlanner` to use `@ClonedPlanningMethod` and updates the README to explain the distinction.  `@PlanningMethod` is changed to inherit from `@ClonedPlanningMethod` but print a warning if it is used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,25 @@ two subclasses:
 
 Each planner has one or more *planning methods*, annotated with either the
 `@LockedPlanningMethod` or `@ClonedPlanningMethod decorator, that look like
-ordinary functions. However, unlike an ordinary function, calling a planning
-method either locks the robot environment or clones the robot's environment
-into a *planning environment* associated with the planner. Planning occurs in
-the locked or cloned environment to allow PrPy to run multiple planners in
-parallel and to paralellize planning and execution.  In general, locked
-planning methods are used for calls that will terminate extremely quickly,
-while cloned planning methods are used for calls that might take a significant
-amount of time.
+ordinary functions. Using these decorators makes other PrPy components
+aware that these methods exist and follow a particular specification that
+allows them to be composed with other PrPy objects automatically.  For
+example, `MetaPlanner`s will report that they can perform planning methods
+that their child motion planners have enumerated via `@PlanningMethod`
+decorators.
+
+`@PlanningMethod` decorators also make sure that calls to planning code are
+executed in a thread-safe manner.  In the case of `@LockedPlanningMethod`,
+this is enforced by locking the calling environment until the planning method
+has completed. In the case of `@ClonedPlanningMethod`, this is enforced by
+cloning the calling environment, and calling the wrapped method with references
+to the cloned environment.  The result of the method is then copied back to the
+calling environment.  `@ClonedPlanningMethod`s can be used to run multiple
+planners in parallel and to parallelize planning and execution.
+
+In general, **locked** planning methods are used for calls that will terminate
+extremely quickly, while **cloned** planning methods are used for calls that
+might take a significant amount of time.
 
 For example, the following code will use OMPL to plan `robot`'s active DOFs
 from their current values to to the `goal_config` configuration:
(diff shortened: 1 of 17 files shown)


README.md
```diff
@@ -25,12 +25,16 @@ two subclasses:
...
+ordinary functions. However, unlike an ordinary function, calling a planning
+method either locks the robot environment or clones the robot's environment
 into a *planning environment* associated with the planner. Planning occurs in
-the cloned environment to allow PrPy to run multiple planners in parallel and
-to paralellize planning and execution.
+the locked or cloned environment to allow PrPy to run multiple planners in
+parallel and to paralellize planning and execution.  In general, locked
+planning methods are used for calls that will terminate extremely quickly,
+while cloned planning methods are used for calls that might take a significant
+amount of time.
```
mkoval: This description does not clearly articulate the difference between the two decorators. For example, it currently implies that `@LockedPlanningMethod` allows PrPy to parallelize multiple planners. I think you should r

mkoval: Typo: paralellize

README.md
```diff
@@ -112,25 +116,25 @@ of these methods accept planner-specific keyword arguments.
...
 Implementing a custom planner requires extending the `BasePlanner` class and
-decorating one or more methods with the `@PlanningMethod` decorator. Extending
-the `BasePlanner` class constructs the planning environment `self.env` and
-allows PrPy to identify your planner as a base planner class, as opposed to a
-meta-planner. The `@PlanningMethod` decorator handles environment cloning and
-allows meta-planners to query the list of planning methods that the planner
-supports (e.g. to generate docstrings).
+decorating one or more methods with the `@LockedPlanningMethod` or
+`@ClonedPlanningMethod` decorator. Extending the `BasePlanner` class constructs
+the planning environment `self.env` and allows PrPy to identify your planner
```
mkoval: What is the purpose of `self.env` in a planner that uses exclusively `@LockedPlanningMethod` decorators? You should note this here.

README.md
```diff
@@ -112,25 +116,25 @@ of these methods accept planner-specific keyword arguments.
...
 Please obey the following guidelines:
 
-- Assume that the cloned environment is locked during the entire call.
-- Subclass constructor **must** call `BasePlanner.__init__`.
-- A `@PlanningMethod` **must not** call another `@PlanningMethod`.
-- Each `@PlanningMethod` **must** accept the first argument `robot`, which is a
+- Assume that the environment is locked during the entire call.
+- Subclass constructors **must** call `BasePlanner.__init__`.
+- Each `@*PlanningMethod` **must** accept the first argument `robot`, which is a
   robot in the cloned environment.
```
mkoval: This is no longer true. The environment will not be cloned when using `@LockedPlanningMethod`.

README.md
```diff
@@ -112,25 +116,25 @@ of these methods accept planner-specific keyword arguments.
...
-- Each `@PlanningMethod` **must** accept `**kwargs` to ignore arguments that
+- Each `@*PlanningMethod` **must** accept `**kwargs` to ignore arguments that
   are not supported by the planner.
-- Each `@PlanningMethod` **must** return a `Trajectory` which was created in
+- Each `@*PlanningMethod` **must** return a `Trajectory` which was created in
   the cloned environment.
```
mkoval: Ditto. This may not be a cloned environment.

mkoval:
I am generally supportive of this change.

My only concern is about the potential confusion introduced by `planner.env`. This environment is created in the `PlannerBase` class, even in sub-class that only use `@LockedPlanningMethod`. We also explicitly clone into `planner.env` inside the `Ranked` meta planner, which is unnecessary for `@LockedPlanningMethod`s.

I am not sure what the best way to address this is. At a minimum, we need to clearly document that `@*PlanningMethod`s must return a `Trajectory` in the same `Environment` as `robot.GetEnv()`. I believe this handles both cases.

psigen:
I can make some doc changes to clarify the above.

I think it's largely harmless to have `PlannerBase` create a single env per planner, because most planners will have cloned methods or a mix of cloned and locked methods, and I would rather not have to divide the `PlannerBase` up and have to switch classes over when they start having `ClonedPlanningMethod`s.

Also, metaplanners will probably need to always use cloning for now, because they need to lock an env to maintain ActiveDOFs, and other initial configs to start up other planners, and there is currently no way to lock an env just until it has been cloned by a planner.

README.md
psigen: Sure I'll fix your typos 😉 
https://github.com/personalrobotics/prpy/commit/854b6f350df115a2c6aea9a4b6bed7a6f5daf18b

mkoval:
:+1: Thanks for improving the documentation. I will merge this once Travis finishes on #278.

mkoval:
@psigen This will not merge cleanly. I suspect that it conflicts with something in #278. Could you resolve this on the command line?

psigen:
Not sure what the issue was, it merged just fine on command-line with no input.

mkoval:
Finally got Travis to run on this. LGTM.

Issue #275, opened 2016-03-16
Why are planning wrappers for external planners in `prpy`?


psigen:
Consider the case of `or_parabolicsmoother`.  We have the `HauserParabolicSmoother` class which lives in `prpy`, but depends on `or_parabolicsmoother`.  Note that since we maintain both packages, we have the ability to put the `prpy` python binding in either one.

It seems more useful to keep the `prpy` binding with the external package (like `or_trajopt`) rather than split (like `or_parabolicsmoother`):

1. **The API for the `prpy` binding is highly dependent on the external package.**  For example, every XML `PlanningParameter` in `or_parabolicsmoother` has to be carefully matched in the `prpy` `Planner` wrapper.  Mismatches cause hard-to-diagnose reverts to default parameters.  This means that `prpy` experiences version churn if _any_ wrapped planner changes API.

2. **Adding an `or_parabolicsmoother.prpy` module is basically free.**  If someone does not use a `prpy` binding included with `or_parabolicsmoother`, they incur no performance penalty or additional runtime dependencies.  It also does not really inflate the size of the package itself.  This is the same logic by which we include `package.xml` and `xxxxConfig.cmake` files in our third-party library repositories.

3. **Unit tests can live with their planners.**  Since bugs in the planners can show up as unit test failures in `prpy` wrappers, it makes some sense that these unit tests should live with the planner implementation.  The same `test_depend` mechanism that we use in `prpy` to include planners can instead be used to include `prpy` in planner tests.

Notes:
* Pure python planners and wrappers for third-party libraries can sensibly live in `prpy`.
* This has nothing to do with the ExternalProject build used in `or_trajopt`, that is an unrelated discussion.

mkoval:
I agree with this as long as, like you say, wrappers for third-party libraries live in `prpy`. My only concern is test re-use. We currently use mixins to run the same tests across multiple planners that implement the same planning methods. We would have to move these mixins to a module that the planner package could implement.

I would be happy to make this change, but I don't have the time or interest in doing so myself. I'd be happy to merge the pull request if you (or someone else) does the legwork.

PR #281, opened 2016-03-26
Perception/chisel

jeking04:
I think we are ready to merge this. @Shushman and I checked that chisel is working. And I have verified simtrack also works with these changes. We should test vncc one more time to ensure I didn't introduce any bugs with by switching to use the ```kinbody_helper``` functions. But in the meantime, lets start gathering comments on this PR and working it through.

### DIFF
--- a/src/prpy/perception/chisel.py
+++ b/src/prpy/perception/chisel.py
@@ -99,9 +99,9 @@ class ChiselModule(PerceptionModule):
             # Apply a transform to the chisel kinbody to put it in the correct
             # location in the OpenRAVE environment
             from kinbody_helper import transform_to_or
-            transform_to_or(detection_frame="/map",
+            transform_to_or(kinbody=chisel_mesh,
+                            detection_frame="/map",
                             destination_frame=self.destination_frame,
-                            reference_link=self.reference_link,
-                            kinbody=chisel_mesh)
+                            reference_link=self.reference_link)
                         
         return chisel_mesh
--- a/src/prpy/perception/chisel.py
+++ b/src/prpy/perception/chisel.py
@@ -40,11 +40,9 @@ robot.DetectObjects()
 mesh_client.SendCommand('GetMesh Chisel/full_mesh')
 '''
 
-
-
 class ChiselModule():
 
-	def __init__(self,mesh_client,service_namespace=None):
+	def __init__(self, env, mesh_client=None, service_namespace=None):
 
 		try:
 			rospy.init_node('chisel_detector',anonymous=True)
@@ -54,13 +52,35 @@ class ChiselModule():
 		if service_namespace is None:
 			service_namespace = 'Chisel'
 
+                if mesh_client is None:
+                        import openravepy
+                        mesh_client = openravepy.RaveCreateSensorSystem(env, 'mesh_marker')
+
 		self.mesh_client = mesh_client
 		self.serv_ns = service_namespace
 
-	def DetectObject(self,robot,**kw_args):
+                from offscreen_render import ros_camera_sim
+                self.camera = ros_camera_sim.RosCameraSim(env)
+                self.camera.start('/head/kinect2/qhd')
+                self.camera_bodies = []
+
+
+                self.detect_chisel_refresh=None
+		
+
+
+	def DetectObject(self, robot, ignored_bodies=None, **kw_args):
     
 		env = robot.GetEnv()
 		already_in_env = False
+
+                if ignored_bodies is None:
+                        ignored_bodies = []
+
+                for b in ignored_bodies:
+                        if b not in self.camera_bodies:
+                                self.camera.add_body(b)
+                                self.camera_bodies.push_back(b)
 
 		#Check if kinbody in env
 		for body in env.GetBodies():
@@ -74,10 +94,12 @@ class ChiselModule():
 			env.RemoveKinBody(chisel_kb)
 
 		#Reset Chisel
-		srv_nm = self.serv_ns+'/Reset'
-		rospy.wait_for_service(srv_nm)
-		detect_chisel_refresh = rospy.ServiceProxy(srv_nm,
-	                                 chisel_msgs.srv.ResetService)
+                if self.detect_chisel_refresh is None:
+                        srv_nm = self.serv_ns+'/Reset'
+                        rospy.wait_for_service(srv_nm)
+                        detect_chisel_refresh = rospy.ServiceProxy(srv_nm,
+                                                                   chisel_msgs.srv.ResetService)
+                detect_chisel_refresh()
 
-	`	#Get Kinbody and load into env
-		#self.mesh_client.SendCommand('GetMesh Chisel/full_mesh')
+		#Get Kinbody and load into env
+		self.mesh_client.SendCommand('GetMesh Chisel/full_mesh')
--- a/src/prpy/perception/chisel.py
+++ b/src/prpy/perception/chisel.py
@@ -112,7 +112,7 @@ class ChiselModule():
                         rospy.Time(),
                         rospy.Duration(10))
                 frame_trans, frame_rot = self.listener.lookupTransform(
-                        '/map', '/herb_base',
+                        '/herb_base', '/map',
                         rospy.Time(0))
                 from tf.transformations import quaternion_matrix
                 offset = numpy.array(numpy.matrix(quaternion_matrix(frame_rot)))
--- a/src/prpy/perception/chisel.py
+++ b/src/prpy/perception/chisel.py
@@ -19,6 +19,19 @@ camera = ros_camera_sim.RosCameraSim(env)
 camera.start('/head/kinect2/qhd')
 camera.add_body(table)
 camera.add_body(glass)
+
+Nodelet manager screen
+rosrun nodelet nodelet manager __name:=nodelet_manager
+Depth converter screen
+rosrun nodelet nodelet load depth_image_proc/convert_metric nodelet_manager image_raw:=/head/kinect2/qhd/image_depth_rect image:=/head/kinect2/qhd/image_depth_rect_float
+In herb2, source herb2_chisel_ws/devel/setup.bash
+Depth mask screen
+roslaunch depth_mask launch_depth_mask.launch (args are correct)
+
+Chisel screen
+roslaunch chisel_ros launch_kinect_local.launch
+
+IN PYTHON
 mesh_client = openravepy.RaveCreateSensorSystem(env, 'mesh_marker')
 chisel_det = ChiselModule(mesh_client,'Chisel')
 
@@ -39,7 +52,7 @@ class ChiselModule():
 			pass
 
 		if service_namespace is None:
-			service_namespace = '/Chisel'
+			service_namespace = 'Chisel'
 
 		self.mesh_client = mesh_client
 		self.serv_ns = service_namespace
@@ -55,16 +68,16 @@ class ChiselModule():
 				already_in_env = True
 				break
 
-	#Remove kinbody if present
+		#Remove kinbody if present
 		if already_in_env:
 			chisel_kb = env.GetKinBody('Chisel/full_mesh')
 			env.RemoveKinBody(chisel_kb)
 
-	#Reset Chisel
+		#Reset Chisel
 		srv_nm = self.serv_ns+'/Reset'
 		rospy.wait_for_service(srv_nm)
 		detect_chisel_refresh = rospy.ServiceProxy(srv_nm,
 	                                 chisel_msgs.srv.ResetService)
 
-	#Get Kinbody and load into env
+	`	#Get Kinbody and load into env
 		#self.mesh_client.SendCommand('GetMesh Chisel/full_mesh')
--- a/src/prpy/perception/chisel.py
+++ b/src/prpy/perception/chisel.py
@@ -1,131 +1,107 @@
-import rospy
-import chisel_msgs.srv
-import chisel_msgs.msg
-import math
-import tf
-import tf.transformations as transformations
-import numpy
-import os.path
+import logging
+from base import PerceptionModule, PerceptionMethod
 
-'''
-Typical way of running
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
-from prpy.perception.chisel import ChiselModule
-from offscreen_render import ros_camera_sim
-robot.DetectObjects()
-table = env.GetKinBody('table127')
-glass = env.GetKinBody('plastic_glass124')
-camera = ros_camera_sim.RosCameraSim(env)
-camera.start('/head/kinect2/qhd')
-camera.add_body(table)
-camera.add_body(glass)
-
-Nodelet manager screen
-rosrun nodelet nodelet manager __name:=nodelet_manager
-Depth converter screen
-rosrun nodelet nodelet load depth_image_proc/convert_metric nodelet_manager image_raw:=/head/kinect2/qhd/image_depth_rect image:=/head/kinect2/qhd/image_depth_rect_float
-In herb2, source herb2_chisel_ws/devel/setup.bash
-Depth mask screen
-roslaunch depth_mask launch_depth_mask.launch (args are correct)
-
-Chisel screen
-roslaunch chisel_ros launch_kinect_local.launch
-
-IN PYTHON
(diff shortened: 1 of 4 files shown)


src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        @param destination_frame The tf frame the detected chisel kinbody should be transformed to
+        (default map)
+        @param reference_link A link on an OpenRAVE kinbody that corresponds to the tf frame
+        specified by the destination_frame parameter. If not provided it is assumed that the 
+        transform from the destination_frame tf frame to the OpenRAVE environment is the identity
+        """
+        import rospy
+
+        try:
+            rospy.init_node('chisel_detector',anonymous=True)
```
mkoval: We shouldn't call `init_node` here because it may only be called once per process. I prefer calling `init_node` only once, in the top-level script, to prevent this from happening twice.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+from base import PerceptionModule, PerceptionMethod
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+class ChiselModule(PerceptionModule):
+    
+    def __init__(self, env, mesh_client=None, service_namespace=None,
+               detection_frame=None, destination_frame=None, reference_link=None):
+        """
```
mkoval: Add a description of what `ChiselModule` does here.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        import rospy
+
+        try:
+            rospy.init_node('chisel_detector',anonymous=True)
+        except rospy.exceptions.ROSException:
+            pass
+
+        if service_namespace is None:
+            service_namespace = 'Chisel'
```
mkoval: Python strings are immutable, so it's fine to move this default into the argument list. Ditto with `map`.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+
+        if mesh_client is None:
+            import openravepy
```
mkoval: Make this a file-level import. It's not like this script would be useful without `openravepy` installed. :smile:

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+            mesh_client = openravepy.RaveCreateSensorSystem(env, 'mesh_marker')
```
mkoval: Raise an exception if this fails (i.e. returns `None`).

mkoval: Also, what package is the `mesh_marker` sensor system from? We should add this as a dependency in the `package.xml` file.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self.mesh_client = mesh_client
+        self.serv_ns = service_namespace
+
+        # Create a ROS camera used to generate a synthetic point cloud
+        #  to mask known kinbodies from the chisel mesh
+        from offscreen_render import ros_camera_sim
```
mkoval: Add this as a dependency in the `package.xml` file.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        self.camera = ros_camera_sim.RosCameraSim(env)
+        self.camera.start('/head/kinect2/qhd')
```
mkoval: What is this topic used for? Isn't this a simulated camera?

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        self.detection_frame = detection_frame
+
+        if destination_frame is None:
+            destination_frame = '/map'
+        self.destination_frame = destination_frame
+        self.reference_link = reference_link
+
+    @PerceptionMethod
+    def DetectObject(self, robot, ignored_bodies=None, **kw_args):
+        """
```
mkoval: Add a `docstring` that explains what this does.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
+import logging
+from base import PerceptionModule, PerceptionMethod
+
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
+class ChiselModule(PerceptionModule):
+    
+    def __init__(self, env, mesh_client=None, service_namespace=None,
+               detection_frame=None, destination_frame=None, reference_link=None):
```
mkoval: What is `detection_frame`? It's missing a docstring.

If this is the camera frame, please replace this with the `frame_id` from the `CameraInfo` message.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        if ignored_bodies is None:
+            ignored_bodies = []
+
+        # Dictionary mapping ignored bodies to start transform
+        orig_bodies = {}
+
+        #Check if chisel kinbody in env
+        already_in_env = False
+        for body in env.GetBodies():
+            if body.GetName() == 'Chisel/full_mesh':
```
mkoval: 1. Where does the name `Chisel/full_mesh` come from?
2. All of this logic could be replaced by `chisel_kb = env.GetKinBody('Chisel/full_mesh')`.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+                already_in_env = True
+                break
+
+        #Remove chisel kinbody if present
+        if already_in_env:
+            chisel_kb = env.GetKinBody('Chisel/full_mesh')
+            env.Remove(chisel_kb)
+
+        try:
+            from kinbody_helper import transform_from_or
```
mkoval: Add this dependency to the `package.xml` file.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+            #   Then we will transform them all back after the server
+            #   performs the detection
+            for b in ignored_bodies:
+                if b not in self.camera_bodies:
+                    with env:
+                        orig_bodies[b] = b.GetTransform()
+                    transform_from_or(kinbody=b, 
+                                      detection_frame=self.detection_frame,
+                                      destination_frame=self.destination_frame,
+                                      reference_link=self.reference_link)
```
mkoval: It looks like the return value from `transform_from_or` is being ignored. What's the purpose of this?

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+            # Add any ignored bodies to the camera
+            # We need to transform all these bodies from their pose
+            #   in OpenRAVE to the equivalent pose in TF so that
+            #   chisel can perform the appropriate masking
+            #   Then we will transform them all back after the server
+            #   performs the detection
+            for b in ignored_bodies:
+                if b not in self.camera_bodies:
+                    with env:
+                        orig_bodies[b] = b.GetTransform()
```
mkoval: I'd prefer to lock around this entire loop *and* the `SendCommand` call. Otherwise, the objects that need to be masked out may move before we do the render.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+                    transform_from_or(kinbody=b, 
+                                      detection_frame=self.detection_frame,
+                                      destination_frame=self.destination_frame,
+                                      reference_link=self.reference_link)
+                    self.camera.add_body(b)
+                    self.camera_bodies.append(b)
```
mkoval: Shouldn't these be cleared before calls?

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+
+            #Reset Chisel
+            if self.reset_detection_srv is None:
```
mkoval: Why is `reset_detection_srv` optional? What would happen if this wasn't available?

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+                import rospy, chisel_msgs.srv
```
mkoval: Add `chisel_msgs` as a `package.xml` file.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+                srv_nm = self.serv_ns+'/Reset'
+                rospy.wait_for_service(srv_nm)
```
mkoval: I suggest passing a `timeout` here. That way we can raise an exception with an informative error message if the Chisel service is not running.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+            #Get Kinbody and load into env
+            self.mesh_client.SendCommand('GetMesh Chisel/full_mesh')
+            chisel_mesh = env.GetKinBody('Chisel/full_mesh')
+
+        finally:
+
+            # Remove any previous bodies in the camera
+            for b in self.camera_bodies:
+                self.camera.remove_body(b)
+            self.camera_bodies = []
```
mkoval: Why are these on `self`? It seems like a local variable would suffice.

src/prpy/perception/chisel.py
```diff
@@ -0,0 +1,139 @@
...
+        already_in_env = False
+        for body in env.GetBodies():
+            if body.GetName() == 'Chisel/full_mesh':
+                already_in_env = True
+                break
+
+        #Remove chisel kinbody if present
+        if already_in_env:
+            chisel_kb = env.GetKinBody('Chisel/full_mesh')
+            env.Remove(chisel_kb)
```
mkoval: It's a bit odd that we create a new `KinBody` with the same name each time we call `DetectObjects`. This will invalidate any references that the user has to the `KinBody`. Could we update the mesh in existing `KinBody`, instead?

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+      that the transform between the OpenRAVE world and the destination_frame 
+      tf frame is the identity)
+    """
+
+    import numpy, tf, rospy
+    listener = tf.TransformListener()
+
+    listener.waitForTransform(
+        detection_frame, destination_frame,
+        rospy.Time(),
```
mkoval: You should use the timestamp from the data used by Chisel here.

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+    @param detection_frame The tf frame the kinbody was originally detected in
+    @param destination_frame A tf frame that has a direct correspondence with
+      a link on an OpenRAVE Kinbody
+    @param reference_link The OpenRAVE link that corresponds to the tf frame
+      given by the destination_frame parameter (if None it is assumed
+      that the transform between the OpenRAVE world and the destination_frame 
+      tf frame is the identity)
+    """
+
+    import numpy, tf, rospy
```
mkoval: Add `tf` as a dependency.

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+    listener = tf.TransformListener()
+
+    listener.waitForTransform(
+        detection_frame, destination_frame,
+        rospy.Time(),
+        rospy.Duration(10))
+
+    frame_trans, frame_rot = listener.lookupTransform(
+        detection_frame, destination_frame,
+        rospy.Time(0))
```
mkoval: Same comments as above.

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+
+    from tf.transformations import quaternion_matrix
+    destination_in_detection = numpy.array(numpy.matrix(quaternion_matrix(frame_rot)))
+    destination_in_detection[:3,3] = frame_trans
+
+    with kinbody.GetEnv():
+        body_in_or = kinbody.GetTransform()
+
+    or_in_destination = numpy.linalg.inv(reference_link.GetTransform())
```
mkoval: Missing lock.

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+    body_in_destination = numpy.dot(or_in_destination, body_in_or)
+
+    body_in_detection = numpy.dot(destination_in_detection, body_in_destination)
+
+    with kinbody.GetEnv():
```
mkoval: You should use the same lock for all of this to avoid synchronization problems.

src/prpy/perception/simtrack.py
```diff
@@ -34,10 +36,13 @@ def __init__(self, kinbody_path, detection_frame, world_frame,
...
             service_namespace='/simtrack'
+        self.service_namespace = service_namespace
 
         self.detection_frame = detection_frame
-        self.world_frame = world_frame
-        self.service_namespace = service_namespace
+        if destination_frame is None:
+            destination_frame='/map'
+        self.destination_frame = destination_frame
+        self.reference_link = reference_link
```
mkoval: Same comments as above about the defaults and `detection_frame`.

src/prpy/perception/kinbody_helper.py
psigen: FYI guys, multiple `import`s on a single line is _not_ PEP8 :neckbeard: 
https://www.python.org/dev/peps/pep-0008/#imports

src/prpy/perception/kinbody_helper.py
```diff
@@ -0,0 +1,87 @@
...
+    @param detection_frame The tf frame the kinbody was originally detected in
+    @param destination_frame A tf frame that has a direct correspondence with
+      a link on an OpenRAVE Kinbody
+    @param reference_link The OpenRAVE link that corresponds to the tf frame
+      given by the destination_frame parameter (if None it is assumed
+      that the transform between the OpenRAVE world and the destination_frame 
+      tf frame is the identity)
+    """
+
+    import numpy, tf, rospy
```
psigen: https://www.python.org/dev/peps/pep-0008/#imports

src/prpy/perception/chisel.py
psigen: https://www.python.org/dev/peps/pep-0008/#imports

mkoval:
This code has a lot of dependencies, including several that are specific to Chisel. Can we move this code to `or_mesh_marker`, which already has those dependencies? We can then make `herbpy` depend on `or_mesh_marker` without adding any dependencies to `prpy`.

This mirrors @psigen's suggestion in #275.

psigen:
I've been doing this for our smoothers. See:
https://github.com/personalrobotics/or_parabolicsmoother/pull/19
and 
https://github.com/personalrobotics/or_circularblender

src/prpy/perception/chisel.py
Shushman: `camera_bodies`? It is the list of bodies that the camera has added since being started, and needs to have state between calls to `DetectObject`

src/prpy/perception/chisel.py
Shushman: The `or_mesh_marker` plugin was designed so as to directly add a `KinBody` to the current `env` and it throws an error if there is an existing one, so we worked around it this way. We can have it like this for now, pending a change to 'or_mesh_marker` which I might do after this.

src/prpy/perception/chisel.py
Shushman: `kinbody_helper` is inside `prpy`.

src/prpy/perception/chisel.py
Shushman: The `kinbody` is being changed inside `transform_from_or`. Is that the right way to do it?

src/prpy/perception/kinbody_helper.py
Shushman: Put lock around entire thing there and in `transform_to_or`

src/prpy/perception/kinbody_helper.py
Shushman: That's not easily available. `Chisel` listens to a bunch of image topics which we don't want to assume the names of. Does it make sense to use `rospy.Time.now()'?

Shushman:
I've gone through `chisel.py` and `kinbody_helper.py` and addressed most to all of @mkoval's initial comments. Could @jeking04 throw another pair of eyes on this, address the comment for `simtrack` and weigh in on a couple of questions I raised?

Issue #293, opened 2016-04-04
env is not defined, in DetectObject() in prpy.perception.apriltags.py


DavidB-CMU:
in `DetectObject()` from `apriltags.py`, env is not defined.

> /home/user/catkin_workspaces/herb_ws/src/prpy/src/prpy/perception/base.pyc
> in __call__(self, instance, robot, *args, **kw_args)
>      12
>      13     def __call__(self, instance, robot, *args, **kw_args):
> ---> 14         return self.func(instance, robot, *args, **kw_args)
>      15
>      16     def __get__(self, instance, instancetype):
> 
> /home/user/catkin_workspaces/herb_ws/src/prpy/src/prpy/perception/apriltags.py
> in DetectObject(self, robot, object_name, **kw_args)
>     145         for obj in added_bodies:
>     146             if obj.GetName() != object_name:
> --> 147                 env.Remove(obj)
>     148                 print "removed object"
>     149             else:
> 
> NameError: global name 'env' is not defined

DavidB-CMU:
Please see PR #294

Issue #292, opened 2016-04-04
DetectObjects() returns duplicate kinbodies in prpy.perception.apriltags.py


DavidB-CMU:
When you call `DetectObjects()` (from `apriltags.py`) for the first time in an environment, 
it returns both *added_kinbodies* and *updated_kinbodies* 

> robot.DetectObjects()
> 
> Out[1]:
> [RaveGetEnvironment(1).GetKinBody('table127'),
>  RaveGetEnvironment(1).GetKinBody('plastic_glass124'),
>  RaveGetEnvironment(1).GetKinBody('table127'),
>  RaveGetEnvironment(1).GetKinBody('plastic_glass124')]

any other subsequent calls return only the updated kinbodies:
> Out[1]:
> [RaveGetEnvironment(1).GetKinBody('table127'),
>  RaveGetEnvironment(1).GetKinBody('plastic_glass124')]

Is this expected? 
I would propose only returning the *added* kinbodies.

DavidB-CMU:
Please see PR #294

DavidB-CMU:
Will not fix in apriltags.py  
This will be fixed by PR https://github.com/personalrobotics/kinbody_detector/pull/4

PR #294, opened 2016-04-04
issue #292 | issue #293

DavidB-CMU:
issue #292 
DetectObjects() returns duplicate kinbodies in prpy.perception.apriltags.py 

 issue #293 
 env is not defined, in DetectObject() in prpy.perception.apriltags.py

### DIFF
--- a/src/prpy/perception/apriltags.py
+++ b/src/prpy/perception/apriltags.py
@@ -134,6 +134,7 @@ class ApriltagsModule(PerceptionModule):
         return_obj = None
         for obj in added_bodies:
             if obj.GetName() != object_name:
+                env = robot.GetEnv()
                 env.Remove(obj)
             else:
                 return_obj = obj


mkoval:
@Shushman @jeking04 Will this change in `DetectObject` behavior break any of our demos?

Shushman:
I don't think so but I'd like to test them out before committing.

jeking04:
I definitely agree with the change to fix #293. But I am not sure I understand the change for #292. It seems to me like the true bug is further down the pipeline: the ```kinbody_detector``` is incorrectly putting those bodies in both the ```added``` and ```updated``` lists. I need a little convincing that the right thing to do is just ignore one list. Either we are incorrectly interpretting the results from the ```kinbody_detector``` or there is a bug in ```kinbody_detector.Update```. I think we need to understand which of these is the true problem and either a) update the code to correctly name and use the two elements returned by the detector or b) fix the bug in ```kinbody_detector```

Shushman:
On a (somewhat) related note, for the `aikido` perception re-write, I am not making a separate `KinBody Detector`, since it seems quite specific to AprilTags. I have just moved the relevant part of the functionality to AprilTags. And it definitely only returns one set of objects - so our future `DetectObjects()` should be fine.

jeking04:
To follow up, I think this is definitely a problem with the ```kinbody_detector```. You can see here:

https://github.com/personalrobotics/kinbody_detector/blob/master/src/kinbody_detector/kinbody_detector.py#L102

Objects added to ```added_kinbodies``` also get put in the ```updated_kinbodies``` list here:

https://github.com/personalrobotics/kinbody_detector/blob/master/src/kinbody_detector/kinbody_detector.py#L107

I propose we remove the change from here and open an issue against the ```kinbody_detector``` repository.

DavidB-CMU:
I discussed this with @jeking04 - I'll take a look at kinbody_detector

DavidB-CMU:
Should we move this block up to the constructor?

>             # TODO: Creating detector is not instant...might want
>             #  to just do this once in the constructor
>             import kinbody_detector.kinbody_detector as kd
>             detector = kd.KinBodyDetector(env,
>                                           marker_data_path,
>                                           kinbody_path,
>                                           marker_topic,
>                                           detection_frame,
>                                           destination_frame,
>                                           reference_link)

DavidB-CMU:
I decided not to move that block into the constructor. There's a lot of parameters that can potentially be overridden.

DavidB-CMU:
I also modified DetectObject() and DetectObjects() to suit the changes to kinbody_detector. 
For example, you can call DetectObject('cup5') and it won't update the pose of the table.

mkoval:
I am fine with merging these changes, and https://github.com/personalrobotics/kinbody_detector/pull/4, if @Shushman and @jeking04 can confirm that this will not break the demo.

DavidB-CMU:
@Shushman and I tested this again this morning. @jeking04 can we merge this in please?

DavidB-CMU:
We decided to strip the latest commit so this PR is finished.

The other issue is now here: https://github.com/personalrobotics/prpy/issues/300

PR #298, opened 2016-04-19
Fixed spurious CloneException errors in planning stack.

psigen:
Due to OpenRAVE weirdness, we are forced to throw a `CloneException` if a robot is cloned from a self-colliding configuration.  We also clone at the beginning of some of our planning operations.

This means that calling a planning method while a robot is in self-collision can create a `CloneException`, which is generally not expected or handled in code that is calling a planning method.  It is also not clear to calling code that this is actually a failure within a planner, and it breaks `MetaPlanner` logic.

In this PR, `ClonedPlanningMethod`s will catch `CloneException`s and wrap them as `ClonedPlanningError`s instead, which inherit from `PlanningError` and are thus handled correctly throughout the planning stack as a form of planning failure.

### DIFF
--- a/src/prpy/clone.py
+++ b/src/prpy/clone.py
@@ -193,7 +193,7 @@ def Cloned(*instances, **kwargs):
         if instance is None:
             clone_instances.append(None)
             continue
-        
+
         # Clone each instance based on its type.
         if isinstance(instance, openravepy.Robot):
             clone_instance = clone_env.GetRobot(instance.GetName())
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -33,10 +33,11 @@ import functools
 import logging
 import numpy
 import openravepy
-from ..clone import Clone
+from ..clone import Clone, CloneException
 from ..futures import defer
 from ..util import CopyTrajectory, GetTrajectoryTags, SetTrajectoryTags
-from .exceptions import PlanningError, UnsupportedPlanningError
+from .exceptions import (ClonedPlanningError, MetaPlanningError,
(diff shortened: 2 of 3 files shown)


mkoval:
I am concerned that letting meta-planners catch `CloneException` as a `PlanningError` will mask deeper issues. All `ClonedPlanningMethod`s will fail if the start configuration is in self-collision. There are only a handful of `LockedPlanningMethod`s, which are all fast heuristic planners, so it is unlikely that any non-trivial planning calls will succeed.

I'd rather fail fast by letting the `CloneException` bubble up to the user, instead of silently ignoring most of the motion planners in a `Sequence`. What is the motivation for this change in behavior?

psigen:
The motivation for this is that _no one_ actually handles `CloneException` anywhere right now.  It's simply not something you expect to happen when making an opaque planning method call.

If I put the robot into collision, and tell it to plan to another place, I don't expect to have to catch a `CloneException` to determine that I gave it an invalid starting location.  I expect a `PlanningError`, because it failed to plan.

Note that the _only_ reason we can get a `CloneException` right now is a situation in which _no_ planner can succeed.  `LockedPlanningMethod`s are also going to fail.  So I see no reason that we shouldn't be handling it the same way as any other error that makes planning fail.

mkoval:
I see your point. I was missing the obvious (in hindsight) fact that a `LockedPlanner` would also fail because *the robot starts in self-collision*. I will merge this once Travis is happy.

psigen:
@mkoval bump.

mkoval:
Travis is happy. Merging.

PR #299, opened 2016-04-20
Features/tracik solver

sjavdani:
Changed the default ik solver for ada to use trac_ik.

Note that I didn't add a dependency to package.xml - I didn't see the nlopt package there. Additionally, or_trac_ik depends on trac-ik which depends on moveit, so forcing everyone to depend on it means everyone has to install all of moveit. So, not sure if you want that.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -32,6 +32,7 @@
   <run_depend>python-rospkg</run_depend>
   <run_depend>python-yaml</run_depend>
   <run_depend>python-lxml</run_depend>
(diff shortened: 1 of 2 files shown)


mkoval:
:+1: I suggest adding `trac-ik` as a dependency of `adapy` so it still gets installed by default.

If no one outside of the lab is using `Mico`, then we should seriusly consider moving `MicoRobot` to `adapy` to avoid situations like this in the future. Any idea if that is the case?

PR #301, opened 2016-04-22
removed mico files

sjavdani:


mkoval:
@sjavdani already ported these files over to `adapy`, so I will merge this once Travis is happy.

PR #302, opened 2016-04-25
PEP8 and lint fixes.

psigen:
This PR cleans up a number of PEP8 and lint issues in `prpy.util` and a few of the `prpy` planners.  There should be no logical changes to the code.

>*Something something deck chairs on the Titanic.* :ship:


mkoval:
Thank you for doing this as a separate PR.

Issue #297, opened 2016-04-14
PlanToConfig doesn't check if robot is already at the config


DavidB-CMU:
prpy.PlanToConfig() doesn't seem to check if the arm is already at the config that you want to go to.

There may also be a problem with wrapping of angles e.g. not realizing a current joint angle matches an angle in the config you want to go too.

mkoval:
There isn't any single place where we could put this logic because each planner provides an independent implementation of `PlanToConfiguration`. Even if it were possible, I am not sure that this is a good idea: most planners naturally cope with this case, i.e. by immediately returning success, without the neeed for an additional check.

May I ask what unexpected behavior you encountered that prompted this issue?

psigen:
@mkoval @DavidB-CMU: I actually encountered a very related issue.

I was planning to a TSR just on the surface of a table.  The TSR bounds in Z had enough slop to go through the table a bit...

What I got was a trajectory that was _half_ a DOF resolution into collision at its goal, which meant that the VanDerCorputGenerator did not check the final goal configuration, instead checking a previous point which was out of collision a half-DOF-resolution step away.

However, when I attempted to plan another trajectory that started at the end of this trajectory, this trajectory was now checked _exactly_ at the start (the end of the previous trajectory), which was in collision, even though the previous planner returned success.  This led to the next planning call failing (and my discovery of bug #298).

This makes me think we should _always_ check the start and goal constraints explicitly: they are boundary conditions that are likely to be used exactly in subsequent steps, so we really want to make sure they _actually_ satisfy constraints like collision and aren't just near another point that satisfies constraints.

mkoval:
This is an important point. I agree your conclusion:

> we should always check the start and goal endpoints' constraints explicitly.

Does this affect any planners besides `SnapPlanner`? Could you put together a pull request that makes the necessary changes?

psigen:
At the moment, `SnapPlanner` is the only planner I've directly observed with this behavior, but we should probably also be on the lookout for similar things in `VectorFieldPlanner`.

DavidB-CMU:
I disagree with that conclusion. 
If you have robot-to-object collision checking problems at the end point, then you can just as easily have an object that may collide at some part way along the trajectory. 
This suggests our calculations for the collision checking resolution are wrong.

You should upload some specific test cases for this. 
I have seen your problem myself, but it was actually related to problems with incorrect collision geometry, incorrect loading of collision geometry, or incorrect cloning of an environment.

[...]

PR #303, opened 2016-04-25
Always check end point in VanDerCorputSampleGenerator()

psigen:
As per discussion in #297, it is valuable to explicitly check the goal state of a planning operation.

This PR addresses the change in `SnapPlanner` by modifying the underlying `VanDerCorputSampleGenerator` to always return the start and end points. There doesn't seem to be a compelling reason for this not to be the case, as the generated samples are aliased over the collision checking interval in an arbitrary fashion anyway.

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1484,6 +1484,10 @@ def VanDerCorputSampleGenerator(start, end, step=2):
     if not (step > 0):
         raise ValueError("The 'step' value must be positive.")
 
+    # Return the start and the end values.
+    yield float(start)
+    yield float(end)
+
     # The duration, rounded to nearest step-size
     mod_end = int(end - (end % step))
     steps_to_take = mod_end / float(step)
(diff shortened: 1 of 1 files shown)


mkoval:
It looks like Travis timed out. I restarted the build.

DavidB-CMU:
@psigen It seems we've gone from a method that did too many collision checks, to a method that is apparently doing not enough collision checks.

Can you add some test cases please?

The result of using `VanDerCorputSampleGenerator` and `SampleTimeGenerator` should be the same, so changes like this should be made and tested with both sequence generators.
.

One extra check won't hurt, but how can the robot can be in a configuration at the end of the trajectory that is within the collision-checking resolution, but yet it collides? Doesn't that mean the resolution is wrong?

Couldn't there be such configurations along the middle of the trajectory?

What do you think @cdellin ?

mkoval:
> But how can the robot can be in a configuration at the end of the trajectory that is within the collision-checking resolution, but yet it collides? Doesn't that mean the resolution is wrong?

This is always the case. We assume that the user selects a collision detection resolution is dense enough to ignore collisions between the sampled points. This is generally accomplished by padding the geometry of everything in the environment by the maximum amount of penetration that may result from the discretization error.

The theoretically sound way of setting padding is to computing the maximum workspace motion of *any point* on the robot caused by changing the robot's configuration by the collision checking resolution. In practice, we arbitrarily choose a value that is "small enough."

This pull request does not change any of that.

> Couldn't there be such configurations along the middle of the trajectory?

The end-point of a trajectory is unique because it is expected to become the *starting point* of future planning queries. Therefore, those queries will fail if the end-point is not feasible. Note that we would have the same problem if we frequently interrupted trajectories during execution, e.g. by dynamically replanning to adapt to a changing environment. That is not something that we are currently exploring, so this pull request is sufficient to fix the immediate issue.

psigen:
I think #304 is a simpler way to get this behavior.  Let's continue discussion there.

PR #304, opened 2016-04-29
Changes VanDerCorputSampleGenerator to include goal.

psigen:
This refactors VanDerCorputSampleGenerator to include the start and end (goal) values as the first two samples to be checked, followed by the discretized intermediate checks within the interval range.

This is an alternate way to address #297 and #303, but avoids the non-termination problem introduced in #303 by simplifying the generator loop.  The resulting output should be identical to the expected output of #303.

**@mkoval @DavidB-CMU let's continue the discussion from #303 here.**

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1544,7 +1544,7 @@ def VanDerCorputSampleGenerator(start, end, step=2):
         if numpy.all(is_checked):
             return
 
-        idx = numpy.digitize((s,), check_bins)
+        idx = numpy.digitize((s,), check_bins, right=True)
         if is_checked[idx]:
             continue
 
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -1510,8 +1510,7 @@ def VanDerCorputSampleGenerator(start, end, step=2):
     Generates a sequence of values from start to end, with specified
     step size, using an approximate binary Van der Corput sequence.
 
-    The end value is also returned if it's more than half the
-    distance from the closest value.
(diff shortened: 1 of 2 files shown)


psigen:
@mkoval: Heads-up that the VDC unit test will fail because it checks the `end` value first instead of last.  I'll fix before merge but I wanted to wait until after we converged on discussion.

psigen:
Ok, since I didn't hear any feedback and the sequences seemed to be pretty correct, I changed the unit tests to match them.

mkoval:
I agree that we explicitly check the endpoint of trajectories inside motion planners. However, I also agree with @DavidB-CMU that this does not belong in our implementation of the Van de Corput sequence - which does not inherently have anything to do with collision checking or trajectories.

I would prefer to either:

1. manually check the endpoint of trajectories in each planner
2. add an optional `check_endpoints` argument to the Van de Corput sequence that defaults to `False`

I am slightly in favor of (2) because it matches the API that @ClintLiddick implemented in https://github.com/personalrobotics/aikido/pull/37.

psigen:
That is why I didn't modify the `VanDerCorputSequence`, I modified the `VanDerCorputSampleGenerator`.  From its docstring:
> This wraps VanDerCorputSequence() in a way that's useful for collision-checking.

This function is NOT the Van Der Corput sequence at all.  It does the relatively arbitrary step-size discretization that we only use for collision checking.

mkoval:
Sorry - I forgot that these were split. This is the right place to add that this functionality. :+1:

PR #311, opened 2016-06-13
Added constrained planning support to OMPLPlanner

jeking04:
This PR adds the ability to do constrained planning with OMPL via the ```pr-constraint``` package developed by @cdellin.

### DIFF
--- a/src/prpy/planning/adapters.py
+++ b/src/prpy/planning/adapters.py
@@ -16,6 +16,9 @@ from ..util import GetManipulatorIndex
 
 
 class PlanToEndEffectorOffsetTSRAdapter(Planner):
+    epsilon = 1e-3
+
+
     """ Plan PlanToEndEffectorOffset using PlanToTSR on a delegate planner.
 
     @param delegate_planner planner that supports goal and constraint TSRs
@@ -24,7 +27,6 @@ class PlanToEndEffectorOffsetTSRAdapter(Planner):
         super(PlanToEndEffectorOffsetTSRAdapter, self).__init__()
 
         self.delegate_planner = delegate_planner
-        self.epsilon = 0.001
         self.env = Environment()
 
 
@@ -43,6 +45,12 @@ class PlanToEndEffectorOffsetTSRAdapter(Planner):
     """
     @ClonedPlanningMethod
     def PlanToEndEffectorOffset(self, robot, direction, distance, **kwargs):
+        chains = self.CreateTSRChains(robot, direction, distance)
+        return self.delegate_planner.PlanToTSR(robot, chains, **kwargs)
+
+
+    @classmethod
+    def CreateTSRChains(cls, robot, direction, distance):
         direction = numpy.array(direction, dtype=float)
 
         if direction.shape != (3,):
@@ -72,13 +80,12 @@ class PlanToEndEffectorOffsetTSRAdapter(Planner):
             T0_w=H_world_w,
             Tw_e=H_w_ee,
             Bw=numpy.array([
-                [-self.epsilon, self.epsilon],
-                [-self.epsilon, self.epsilon],
+                [-cls.epsilon, cls.epsilon],
+                [-cls.epsilon, cls.epsilon],
                 [min(0.0, distance), max(0.0, distance)],
-                [-self.epsilon, self.epsilon],
-                [-self.epsilon, self.epsilon],
-                [-self.epsilon, self.epsilon]]),
+                [-cls.epsilon, cls.epsilon],
+                [-cls.epsilon, cls.epsilon],
+                [-cls.epsilon, cls.epsilon]]),
             manip=manip_index))
 
-        return self.delegate_planner.PlanToTSR(
-            robot, [goal_chain, constraint_chain], **kwargs)
+        return [goal_chain, constraint_chain]
--- a/src/prpy/planning/adapters.py
+++ b/src/prpy/planning/adapters.py
@@ -1,0 +1,84 @@
+import numpy
+from openravepy import Environment
+from .base import (
+    ClonedPlanningMethod,
+    Planner,
+)
+from ..kin import (
+    H_from_op_diff,
+    invert_H,
+)
+from ..tsr.tsr import (
+    TSR,
+    TSRChain,
+)
+from ..util import GetManipulatorIndex
+
+
+class PlanToEndEffectorOffsetTSRAdapter(Planner):
+    """ Plan PlanToEndEffectorOffset using PlanToTSR on a delegate planner.
+
+    @param delegate_planner planner that supports goal and constraint TSRs
+    """
+    def __init__(self, delegate_planner):
+        super(PlanToEndEffectorOffsetTSRAdapter, self).__init__()
+
+        self.delegate_planner = delegate_planner
+        self.epsilon = 0.001
+        self.env = Environment()
+
+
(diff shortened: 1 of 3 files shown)


src/prpy/planning/ompl.py
```diff
@@ -245,3 +246,100 @@ def ShortcutPath(self, robot, path, timeout=1., **kwargs):
...
+        @param robot
+        @param tsrchains A list of TSRChain objects to respect during planning
+        @param planner_range the max stepsize for the planner before a projection
+         is applied
+        @param ompl_args ompl RRTConnect specific parameters
+        @return traj
+        """
+        if ompl_args is None:
+            ompl_args = {}
+        ompl_args['range'] = planner_range
```
mkoval: Why do we manually set `range` only for constrained planners?

Also, I think the default value should be derived from DOF resolution. We have some logic to do this in the `RRTConnectPlanner` wrapper class.

src/prpy/planning/ompl.py
```diff
@@ -245,3 +246,100 @@ def ShortcutPath(self, robot, path, timeout=1., **kwargs):
...
+                          manip = robot.GetActiveManipulatorIndex())
+
+            traj_tsr_chain = TSRChain(constrain=True, TSRs=[trajtsr])
+
+            # Compute a goal pose
+            goal_pose = manip.GetEndEffectorTransform()
+            goal_pose[:3,3] += direction*distance/numpy.linalg.norm(direction)
+
+            # Get all IKs for the goal pose and send them all to the planner
+            goal_configs = manip.FindIKSolutions(goal_pose, 0)
```
mkoval: I thought `or_ompl` supported goal sets. Why do we manually find IK solutions here?

src/prpy/planning/ompl.py
cdellin: From what I remember (and I'll have to take a closer look to determine this more closely), the goal here was to emulate the way CBiRRT places vertices in its RRT after each projection.  Since the default `range` parameter is rather large, we currently overwrite it only for constrained planners.

mkoval:
This is ready for review. @jeking04 can you take a look?

This depends on https://github.com/personalrobotics/or_ompl/pull/72.

src/prpy/planning/adapters.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+class PlanToEndEffectorOffsetTSRAdapter(Planner):
+    epsilon = 1e-3
+
+
+    """ Plan PlanToEndEffectorOffset using PlanToTSR on a delegate planner.
+
+    @param delegate_planner planner that supports goal and constraint TSRs
+    """
```
gilwoolee: This docstring needs to be under method declaration

src/prpy/planning/adapters.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, delegate_planner):
+        super(PlanToEndEffectorOffsetTSRAdapter, self).__init__()
+
+        self.delegate_planner = delegate_planner
+        self.env = Environment()
+
+
+    """ Plan to a desired end-effector offset with fixed orientation.
```
gilwoolee: This docstring needs to be under method declaration

src/prpy/planning/ompl.py
```diff
@@ -81,7 +81,7 @@ class OMPLPlanner(BasePlanner):
     @param ompl_args dictionary of planner-specific arguments
```
gilwoolee: Add docstring for `support_constraints`

src/prpy/planning/ompl.py
```diff
@@ -175,12 +176,22 @@ def _Plan(self, robot, goal=None, tsrchains=None, timelimit=None,
...
+                        'OMPL does not support start TSRs.')
+
+                if chain.sample_goal:
+                    extraParams += '<{k:s}>{v:s}</{k:s}>'.format(
+                        k='tsr_chain', v=SerializeTSRChain(chain))
+
+                if chain.constrain:
+                    if self.supports_constraints:
+                        extraParams += '<{k:s}>{v:s}</{k:s}>'.format(
+                            k='tsr_chain_constraint', v=chain.to_json())
```
gilwoolee: How come this uses `chain.to_json` and above one use `SerializeTSRChain`?

src/prpy/planning/cbirrt.py
```diff
@@ -102,76 +107,30 @@ def PlanToEndEffectorPose(self, robot, goal_pose, **kw_args):
...
-                              [-epsilon,            epsilon]])
-
-            traj_tsr = prpy.tsr.tsr.TSR(
-                T0_w=H_world_w, Tw_e=H_w_ee, Bw=Bw,
-                manip=robot.GetActiveManipulatorIndex())
-            traj_tsr_chain = prpy.tsr.tsr.TSRChain(constrain=True,
-                                                   TSRs=[traj_tsr])
+        chains = PlanToEndEffectorOffsetTSRAdapter.CreateTSRChains(
+                    robot, direction, distance)
+        return self.PlanToTSR(robot, tsr_chains=chains, **kw_args)
```
gilwoolee: The previous version had `kw_args.setdefault('psample', 0.1)`. Is this no longer necessary?

mkoval: This function now calls `PlanToTSR`, which has the same `setdefault('psample', 0.1)` line in it.

src/prpy/planning/ompl.py
mkoval: I added a comment to address this.

src/prpy/planning/ompl.py
```diff
@@ -175,12 +180,26 @@ def _Plan(self, robot, goal=None, tsrchains=None, timelimit=None,
...
+                    extraParams += '<{k:s}>{v:s}</{k:s}>'.format(
+                        k='tsr_chain', v=SerializeTSRChain(chain))
+
+                # Constraint TSRs are parsed by pr_constraint_or_ompl, which
+                # uses a JSON format. This differs from the space delimited
+                # format used by CBiRRT.
+                if chain.constrain:
+                    if self.supports_constraints:
+                        extraParams += '<{k:s}>{v:s}</{k:s}>'.format(
+                            k='tsr_chain_constraint', v=chain.to_json())
```
gilwoolee: Why does this use `chain.to_json` and two lines above it's `SerializeTSRChain`?

mkoval: It's just how it is. `or_ompl` (which uses the CBiRRT format) deserializes goal TSRs and `pr_constraint_or_ompl` (which uses JSON) deserializes constraint TSRs.

psigen: Can we port or_ompl to use the json format as well?

mkoval: I am fine with adding the JSON format as an option. That change belongs in a separate pull request.

If we do so, I would like to keep `<tsr_chain>` using the CBiRRT format for backwards compatibility.

mkoval: I created an issue to track this here: https://github.com/personalrobotics/or_ompl/issues/73.

mkoval:
@gilwoolee signed off on this so I am merging it.

src/prpy/planning/ompl.py
psigen: Sounds good. Thanks.

PR #313, opened 2016-07-02
Feature/tsr gripper library

athackst:
adding support for generic robots

### DIFF
--- a/src/prpy/__init__.py
+++ b/src/prpy/__init__.py
@@ -28,7 +28,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import base, controllers, dependency_manager, logger, ik_ranking, planning, perception, simulation, tsr, viz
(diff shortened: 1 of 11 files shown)


athackst:
@mkoval @jeking04 I don't think this will have any adverse affects to current prpy operation, so we could go ahead and merge it in.

src/prpy/__init__.py
```diff
@@ -28,7 +28,7 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-import base, dependency_manager, logger, ik_ranking, planning, perception, simulation, tsr, viz
+import base, controllers, dependency_manager, logger, ik_ranking, planning, perception, rave, simulation, tsr, viz
```
mkoval: Please remove this import - it will cause a regression of #308.

mkoval:
This looks good to me except for the one minor comment above.

@jeking04 Can you also confirm that this won't break anything?

jeking04:
👍  looks good. this is useful additional functionality.

athackst:
updated

PR #315, opened 2016-07-16
Fix CHOMP DistanceFieldManager field generation to properly disable other kinbodies

cdellin:
When the CHOMP module computes a distance field attached to a kinbody, it does not internally disable other kinbodies during collision checking (this enables multiple bodies to be included in one field, which you often want).  However, the `DistanceFieldManager` assumes that there is one field per kinbody ... but when generating that field, it was not properly disabling other kinbodies.  This led to incorrect fields and poor CHOMP performance in `prpy`.  This PR fixes this bug.

### DIFF
--- a/src/prpy/planning/chomp.py
+++ b/src/prpy/planning/chomp.py
@@ -90,10 +90,8 @@ class DistanceFieldManager(object):
                     for other_body in self.env.GetBodies():
                         if other_body == body:
                             continue
-                        if not other_body.IsEnabled():
-                            continue
                         other_bodies.append(other_body)
-                    other_savers = map(openravepy.KinBodyStateSaver, other_bodies)
+                    other_savers = [openravepy.KinBodyStateSaver(body,openravepy.KinBody.SaveParameters.LinkEnable) for body in other_bodies]
                     with contextlib.nested(*other_savers):
                         for other_body in other_bodies:
                             other_body.Enable(False)
--- a/src/prpy/planning/chomp.py
+++ b/src/prpy/planning/chomp.py
@@ -91,7 +91,7 @@ class DistanceFieldManager(object):
                         if other_body == body:
                             continue
                         other_bodies.append(other_body)
-                    other_savers = [openravepy.KinBodyStateSaver(body,openravepy.KinBody.SaveParameters.LinkEnable) for body in other_bodies]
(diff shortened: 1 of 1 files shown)


src/prpy/planning/chomp.py
```diff
@@ -84,7 +85,20 @@ def sync(self, robot):
...
-                    self.module.computedistancefield(body, cache_filename=cache_path)
+                    # Temporarily disable other bodies for distance field computation
+                    other_bodies = []
+                    for other_body in self.env.GetBodies():
+                        if other_body == body:
+                            continue
+                        if not other_body.IsEnabled():
+                            continue
+                        other_bodies.append(other_body)
+                    other_savers = map(openravepy.KinBodyStateSaver, other_bodies)
```
mkoval: Could you explicitly pass `Save_LinkEnable` in here?

src/prpy/planning/chomp.py
```diff
@@ -84,7 +85,20 @@ def sync(self, robot):
                         body_name, os.path.basename(cache_path)
                     )
 
-                    self.module.computedistancefield(body, cache_filename=cache_path)
+                    # Temporarily disable other bodies for distance field computation
+                    other_bodies = []
+                    for other_body in self.env.GetBodies():
+                        if other_body == body:
+                            continue
+                        if not other_body.IsEnabled():
```
mkoval: What does `body.IsEnabled()` return if the body has a mixture of enabled and disabled links? It might be safer to remove this check and disable everything.

src/prpy/planning/chomp.py
cdellin: `Save_LinkEnable` is set by default [0], but I changed this anyways for clarity (-:

[0] http://openrave.org/docs/latest_stable/coreapihtml/classOpenRAVE_1_1KinBody_1_1KinBodyStateSaver.html#a97962441c40bb5e3ff13a385f7a84f75

src/prpy/planning/chomp.py
cdellin: `body.IsEnabled()` returns `True` if any link is enabled [0] so this check does the right thing, but I removed the check for simplicity.

[0] http://openrave.org/docs/latest_stable/coreapihtml/classOpenRAVE_1_1KinBody.html#a29a5ffa80c81125b8d5ccb40e4fe44be

cdellin:
Comments addressed!

PR #321, opened 2016-07-22
Making TSRPlanner more efficient

gilwoolee:
@cdellin and @mkoval noticed that there was 2 second delay in all `TSRPlanner.PlanToTSR` calls. This is because it sampled TSRs and computed ik solutions with collision checking  until `tsr_timeout`. It then ranked the solutions and batch-ran PlanToConfiguration. This ended up taking `tsr_timeout` time on all `PlanToTSR` calls, even when easy solution existed.

We changed this to sample and compute ik solutions until `num_samples` are sampled or `tsr_timeout` is reached, whichever happens first, without checking collision. Then we rank the ik solutions and lazily create batches with collision checking, and run `PlanToConfiguration` until solution is found or `num_attempts` is reached. We also reduced default `tsr_timeout` from 2.0s to 0.5s.

### DIFF
--- a/src/prpy/planning/tsr.py
+++ b/src/prpy/planning/tsr.py
@@ -145,7 +145,6 @@ class TSRPlanner(BasePlanner):
 
         # Configure the robot to use the active manipulator for planning.
         from openravepy import CollisionOptions, CollisionOptionsStateSaver
-
         p = openravepy.KinBody.SaveParameters
         with robot.CreateRobotStateSaver(p.ActiveDOF),\
               CollisionOptionsStateSaver(self.env.GetCollisionChecker(), 
@@ -153,7 +152,11 @@ class TSRPlanner(BasePlanner):
             
             robot.SetActiveDOFs(manipulator.GetArmIndices())
 
+            # Attempt num_attempts planning attempts
             for i_attempt in xrange(num_attempts):
+
+                # Build a set of the next chunk_size collision-free
+                # IK solutions (considered in ranked order).
                 ik_set = []
                 while len(ik_set) < chunk_size and ranked_ik_solutions:
                     ik_solution = ranked_ik_solutions.pop(0)
@@ -162,7 +165,6 @@ class TSRPlanner(BasePlanner):
                         ik_set.append(ik_solution)
 
                 # Try planning to each solution set in descending cost order.
-                # for i, ik_set in ik_set_list:
                 try:
                     if len(ik_set) > 1: 
                         traj = delegate_planner.PlanToConfigurations(
--- a/src/prpy/planning/tsr.py
+++ b/src/prpy/planning/tsr.py
@@ -62,9 +62,9 @@ class TSRPlanner(BasePlanner):
     """
 
     @ClonedPlanningMethod
-    def PlanToTSR(self, robot, tsrchains, tsr_timeout=2.0,
+    def PlanToTSR(self, robot, tsrchains, tsr_timeout=0.5,
                   num_attempts=3, chunk_size=1, ranker=None,
-                  max_deviation=2 * numpy.pi, **kw_args):
+                  max_deviation=2 * numpy.pi, num_samples=10, **kw_args):
         """
         Plan to a desired TSR set using a-priori goal sampling.  This planner
         samples a fixed number of goals from the specified TSRs up-front, then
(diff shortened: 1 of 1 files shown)


cdellin:
I must say, this is surprisingly well-written code!

mkoval:
I agree. Whoever (singular or plural) wrote this did an *amazing* job. :grin:

src/prpy/planning/tsr.py
```diff
@@ -122,7 +122,7 @@ def PlanToTSR(self, robot, tsrchains, tsr_timeout=2.0,
             ik_param = IkParameterization(
                 tsrchain.sample(), IkParameterizationType.Transform6D)
             ik_solution = manipulator.FindIKSolutions(
-                ik_param, IkFilterOptions.CheckEnvCollisions,
+                ik_param, IkFilterOptions.IgnoreSelfCollisions,
```
psigen: Why do we want to ignore self collisions here?

src/prpy/planning/tsr.py
```diff
@@ -141,41 +141,46 @@ def PlanToTSR(self, robot, tsrchains, tsr_timeout=2.0,
...
+        ranked_ik_solutions = list(ranked_ik_solutions)
 
         # Configure the robot to use the active manipulator for planning.
+        from openravepy import CollisionOptions, CollisionOptionsStateSaver
+
         p = openravepy.KinBody.SaveParameters
-        with robot.CreateRobotStateSaver(p.ActiveDOF):
+        with robot.CreateRobotStateSaver(p.ActiveDOF),\
+              CollisionOptionsStateSaver(self.env.GetCollisionChecker(), 
+                                         CollisionOptions.ActiveDOFs):
```
psigen: @mkoval is there a way to wrap this line with a paren instead of a `\`?

src/prpy/planning/tsr.py
psigen: Oh, is this because collision checking is expensive and we are deferring it?

cdellin: Correct!

src/prpy/planning/tsr.py
cdellin: Unfortunately parens get interpreted as a tuple, and it doesn't work (that was the first thing we tried).  See http://stackoverflow.com/a/31571166.

src/prpy/planning/tsr.py
```diff
@@ -141,41 +141,46 @@ def PlanToTSR(self, robot, tsrchains, tsr_timeout=2.0,
...
+            
             robot.SetActiveDOFs(manipulator.GetArmIndices())
 
-            # Try planning to each solution set in descending cost order.
-            for i, ik_set in ik_set_list:
+            for i_attempt in xrange(num_attempts):
+                ik_set = []
+                while len(ik_set) < chunk_size and ranked_ik_solutions:
```
psigen: Can we add a comment explaining that this loop tries to accumulate up to the next `chunk_size` of `ranked_ik_solutions` that are not in collision?

cdellin: Addressed!

cdellin:
That's a very good point! I wonder if we can make a test case that provokes that.

I can imagine a way to address this which introduces an additional generator of batches of TSR pose sample.  Each batch would be treated the way we do here (generate all IK solutions for each pose sample, then rank all solutions, then lazily collision check them in ranked order).  If processing that batch terminates without `num_attempts` planning calls being performed, then a new batch of pose samples can be considered. I think that would approximately replicate the current behavior.  (Of course, it would also introduce another parameter, `num_tsrsample_batches` or somesuch.

I would imagine that implementing this behaviour would entail either another outer loop, or rewriting all of this to use fancy generators and `itertools`.  @pkv, @mkoval, and @gilwoolee, what do y'all think?

gilwoolee:
One thing to note is that it's not picking 10 IKs. it's picking 10 tsr samples and stacking all IKsolutions from the 10 tsr samples before ranking them. (It is already sort of doing what @cdellin  suggested, except it ranks all IKs together.) So it may have more than 10 IK solutions, among which it starts with the "best" IK solution batch. I wonder how likely it is that the first `num_attempts` batch of IK solutions from uniformly sampled 10 tsr samples will fail. Maybe increasing `num_attempts` might be more important if we think that it's worth trying more IKs.

gilwoolee:
Hmm. I don't know what's causing this, but with this branch and master of table_clearing it gets stuck in executing the full table_clearing run ( `rosrun table_clearing run.py --planner restart --nobowl --noglass --perception-sim`). 
It plans fine, but at the time of execution I get : 
```
openravepy.interfaces.BaseManipulation: __del__, failed to lock environment for BaseManipulation.__del__!
[WARNING] [openravepy.interfaces.BaseManipulation:BaseManipulation.py:47]:__del__: failed to lock environment for BaseManipulation.__del__!
```
I'll investigate this, so please not merge before fixing this.

mkoval:
I took a stab at addressing those concerns in #322.

cdellin:
OK, I found a workaround for the bug that @gilwoolee discovered last night. (Issue about potentially deeper issue in #324). Gilwoo, can you try the latest version of this and confirm that you no longer see that weird locking issue?

mkoval:
Closing this in favor of #322.

PR #325, opened 2016-07-23
Add metadata about determinism to the planning pipeline

mkoval:
This pull request adds three new features to the PrPy planning pipeline:

- **`DETERMINISTIC_TRAJECTORY` Trajectory Tag:** Indicates whether repeating the same planning query will produce the same trajectory
- **`DETERMINSTIC_ENDPOINT` Trajectory Tag:** Indicates whether repeating the same planning query will result in a trajectory with the same endpoint as this trajectory.
- **`determinstic` flag on `PlanningError`:** Indicates whether repeating the same planning query could result in success.

The trajectory tags default to being absent and the `deterministic` flag defaults to `None`. I strongly suggest *logging a warning* and **not** silently defaulting to `False` if they are not set. It is possible that I missed a few calls in the PrPy motion planners that we need to annotate.

This data metadata is correctly propagated by the `Sequence`, `MethodMask`, and `FirstSupported` `MetaPlanner`s. Note that the behavior of `Sequence` is a bit unintuitive: it will flag a trajectory as non-deterministic if _any previous planner in the sequence_ was non-determinstic.

There are a few additional points of subtly:

- **Seeds.** I consider the value of a random number generator seed, even if it is passed explicitly as an argument, to be non-deterministic. This is the only useful way to model randomness: _almost everything_ on a computer deterministic if we consider PRNGs to be deterministic.
- **`DETERMINSTIC_ENDPOINT` on `Sequence`.** I conservatively flag the endpoint of a trajectory as non-deterministic if a previous failure was non-deterministic. This is overly conservative for `PlanToConfiguration`, where the endpoint is _always_ deterministic.
- **Trajopt.** We need to port these changes over to `or_trajopt` separately since it lives in a separate repository.

### DIFF
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -360,7 +360,8 @@ class Sequence(MetaPlanner):
 
                 errors[planner] = e
 
-        raise MetaPlanningError('All planners failed.', errors)
+        raise MetaPlanningError(
+            'All planners failed.', errors, deterministic=is_sequence_deterministic)
 
 
 class Ranked(MetaPlanner):
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -284,9 +284,15 @@ class MetaPlanner(Planner):
 
 
 class Sequence(MetaPlanner):
-    def __init__(self, *planners):
+    KNOWN_KWARGS = set(['allow_nondeterministic'])
+
+    def __init__(self, *planners, **kwargs):
+        assert self.KNOWN_KWARGS.issuperset(kwargs.keys())
+
         super(Sequence, self).__init__()
         self._planners = planners
+        self._allow_nondeterministic = kwargs.get(
+            'allow_nondeterministic', False)
 
     def __str__(self):
         return 'Sequence({:s})'.format(', '.join(map(str, self._planners)))
@@ -299,8 +305,11 @@ class Sequence(MetaPlanner):
         from ..util import Timer
 
         errors = dict()
+        is_sequence_deterministic = True
 
         for planner in self._planners:
+            e = None
+
             try:
                 if planner.has_planning_method(method):
                     logger.info('Sequence - Calling planner "%s".', str(planner))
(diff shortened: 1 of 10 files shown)


src/prpy/planning/base.py
```diff
@@ -298,6 +318,23 @@ def plan(self, method, args, kw_args):
...
+                        }, append=True)
+
+                        if not self._allow_nondeterministic:
+                            logger.warning(
+                                'Tagging trajectory as non-deterministic because an'
+                                ' earlier planner in the Sequence threw a'
+                                ' non-deterministic PlanningError. Pass the'
+                                ' "allow_nondeterministic" to this Sequence'
+                                ' constructor if you intended this behavior.')
+
```
gilwoolee: I'm a bit confused by `_allow_nondeterministic` flag. Why do we need this? You set the tags to `deterministic_traj = False, deterministic_endpoint = False` even if `self._allow_nondeterministic=False`.

src/prpy/planning/base.py
```diff
@@ -298,6 +318,23 @@ def plan(self, method, args, kw_args):
                     with Timer() as timer:
                         output = planner_method(*args, **kw_args)
 
+                    if not is_sequence_deterministic:
+                        # TODO: It is overly conservative to set _ENDPOINT,
+                        # e.g. for PlanToConfiguration. Unfortunately, there is
+                        # no easy way to detect this special case.
+                        SetTrajectoryTags(output, {
+                            Tags.DETERMINISTIC_TRAJECTORY: False,
+                            Tags.DETERMINISTIC_ENDPOINT: False,
```
gilwoolee: Can we check the existing `Tags.DETERMINISTIC_ENDPOINT` of `output`? `PlanToConfiguration` should set this to `True` before returning a trajectory.

src/prpy/planning/base.py
mkoval: It's an option to suppress the warning. It's technically valid to put a deterministic planner after a non-deterministic planner, but it's generally a bad idea because you may end up re-running the deterministic planner on the same problem again.

src/prpy/planning/base.py
mkoval: This condition is triggered by _previous_ plans in the `Sequence` failing. We would have to check whether all previous planners that failed _would have_ had a deterministic endpoint _had they succeeded_.

We could theoretically implement that by adding an `endpoint_deterministic` flag to `PlanningError`. However, it's a bit too counterfactual for my taste.

mkoval:
This now needs to be merged after #334. I had to fix that regression to test this code.

Issue #243, opened 2015-12-11
Robot deserialization - IK issue


jeking04:
Currently we can serialize and deserialize a robot through the prpy.serialization interface. However, the ik database is not properly loaded/initialized when a robot is deserialized. 

NOTE: Current workaround is to load the robot separately, then use deserialization to set its pose/active manipulator/etc... properties.

mkoval:
This is partially addressed by @cdellin  in #328.

PR #328, opened 2016-07-25
Add optional uri_only serialization/deserialization support, logged planner, and replay-log script

cdellin:
This adds support for the `uri_only` flag to the PrPy serialization functions (defaults to `False`).  When this flag is set, the fixed aspects (e.g. geometry and kinematics) of kinbodies and robots in the scene that have URIs (e.g. OpenRAVE xml paths or URDF/SRDF file paths) are not serialized, and instead only the URI itself is serialized.  This is an immediate workaround for a bug in the serialization functions (or in OpenRAVE) #243 which precludes roundtripping HERB with the correct kinematics hashes, resulting in busted IK solvers.  Other advantages include reducing the size of the serialized files, and allowing planning requests to be replayed while robot or object models are modified.

This PR also contains a `Logged` planner, which takes a delegate planner, and logs both the incoming planning request and some simple data about the result, to a file.  There is also a corresponding `replay-log.py` script which allows the planning requests in these log files to be replayed.  (It is expected that downstream users, such ash HerbPy, may want to provide their own log replay facilities which can make use of robot-specific helpers and planners.)

We've been using this a bit for ISER, and it appears to be working well, so I think it's ready for a review.

### DIFF
--- a/src/prpy/planning/logged.py
+++ b/src/prpy/planning/logged.py
@@ -62,9 +62,10 @@ class LoggedPlanner(prpy.planning.base.MetaPlanner):
      - log-20160101-010203.567890.yaml
     """
 
-    def __init__(self, planner):
+    def __init__(self, planner, uri_only=True):
         super(LoggedPlanner, self).__init__()
         self.planner = planner
+        self.uri_only = uri_only
     
     def __str__(self):
         return 'Logged({0:s})'.format(self.planner)
@@ -96,7 +97,7 @@ class LoggedPlanner(prpy.planning.base.MetaPlanner):
             raise RuntimeError('could not retrieve env from planning request!')
         
         # serialize environment
-        envdict = prpy.serialization.serialize_environment(env, uri_only=True)
+        envdict = prpy.serialization.serialize_environment(env, uri_only=self.uri_only)
         
         # serialize planning request
         reqdict = {}
@@ -129,7 +130,7 @@ class LoggedPlanner(prpy.planning.base.MetaPlanner):
         
         except:
             resdict['ok'] = False
-            resdict['exception'] = 'unknown exception type'
+            resdict['exception'] = 'Unknown exception type: {}'.format(repr(sys.exc_info()[1]))
             raise
         
         finally:
--- a/src/prpy/planning/logged.py
+++ b/src/prpy/planning/logged.py
@@ -1,3 +1,33 @@
+#!/usr/bin/env python
+
+# Copyright (c) 2016, Carnegie Mellon University
+# All rights reserved.
+# Authors: Chris Dellin <cdellin@gmail.com>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of Carnegie Mellon University nor the names of its
+#   contributors may be used to endorse or promote products derived from this
+#   software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
 from __future__ import print_function
 from __future__ import unicode_literals
 from __future__ import absolute_import
@@ -13,8 +43,25 @@ import yaml
 import prpy.planning.base
 import prpy.serialization
 
+def serialize_or_errorstr(value):
+    try:
+        return prpy.serialization.serialize(value)
+    except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+        return 'Error: {}'.format(str(ex))
 
 class LoggedPlanner(prpy.planning.base.MetaPlanner):
+    """Planner wrapper which serializes planning requets and responses.
+
+    LoggedPlanner delegates all planning calls to the sub-planner
+    passed to it on construction.  It does not affect the input to
+    or output from this planner.
+
+    On every request, LoggedPlanner serializes the environment,
+    planning method request arguments, and the returned trajectory.
+    It creates a timestamped yaml file in the current directory:
+     - log-20160101-010203.567890.yaml
+    """
+
     def __init__(self, planner):
         super(LoggedPlanner, self).__init__()
         self.planner = planner
@@ -56,17 +103,12 @@ class LoggedPlanner(prpy.planning.base.MetaPlanner):
         reqdict['method'] = method # string
         reqdict['args'] = []
         for arg in args:
-            try:
-                reqdict['args'].append(prpy.serialization.serialize(arg))
-            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
-                reqdict['args'].append(str(ex))
+            reqdict['args'].append(serialize_or_errorstr(arg))
         reqdict['kw_args'] = {}
         for key,value in kw_args.items():
+            # if key serialization fails, it thros UnsupportedTypeSerializationException
             key = prpy.serialization.serialize(key)
-            try:
-                reqdict['kw_args'][key] = prpy.serialization.serialize(value)
-            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
-                reqdict['kw_args'][key] = 'Error: ' + str(ex)
+            reqdict['kw_args'][key] = serialize_or_errorstr(value)
         
         # get ready to serialize result
         resdict = {}
@@ -77,6 +119,7 @@ class LoggedPlanner(prpy.planning.base.MetaPlanner):
             resdict['ok'] = True
             resdict['traj_first'] = list(map(float,traj.GetWaypoint(0)))
             resdict['traj_last'] = list(map(float,traj.GetWaypoint(traj.GetNumWaypoints()-1)))
+            resdict['traj'] = traj.serialize(0)
             return traj
             
         except prpy.planning.PlanningError as ex:
--- a/src/prpy/planning/logged.py
+++ b/src/prpy/planning/logged.py
@@ -1,0 +1,102 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+from __future__ import absolute_import
+from __future__ import division
+import collections
+import math
+import time
+
+import openravepy
+import numpy
+import yaml
+
+import prpy.planning.base
+import prpy.serialization
+
+
+class LoggedPlanner(prpy.planning.base.MetaPlanner):
+    def __init__(self, planner):
+        super(LoggedPlanner, self).__init__()
+        self.planner = planner
+    
+    def __str__(self):
+        return 'Logged({0:s})'.format(self.planner)
+    
+    def has_planning_method(self, method_name):
+        return self.planner.has_planning_method(method_name)
+    
+    def get_planning_method_names(self):
+        return self.planner.get_planning_method_names()
+    
+    def get_planners(self, method_name):
+        return [self.planner]
+    
+    def plan(self, method, args, kw_args):
+        
+        # get log file name
+        stamp = time.time()
+        struct = time.localtime(stamp)
+        fn = 'log-{}.{:06d}.yaml'.format(
+            time.strftime('%Y%m%d-%H%M%S', struct),
+            int(1.0e6*(stamp-math.floor(stamp))))
+        
+        # retrieve enviroment
+        if 'robot' in kw_args:
+            env = kw_args['robot'].GetEnv()
+        elif len(args) > 0:
+            env = args[0].GetEnv()
+        else:
+            raise RuntimeError('could not retrieve env from planning request!')
+        
+        # serialize environment
+        envdict = prpy.serialization.serialize_environment(env, uri_only=True)
+        
+        # serialize planning request
+        reqdict = {}
+        reqdict['method'] = method # string
+        reqdict['args'] = []
+        for arg in args:
+            try:
+                reqdict['args'].append(prpy.serialization.serialize(arg))
+            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+                reqdict['args'].append(str(ex))
+        reqdict['kw_args'] = {}
+        for key,value in kw_args.items():
+            key = prpy.serialization.serialize(key)
+            try:
+                reqdict['kw_args'][key] = prpy.serialization.serialize(value)
+            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+                reqdict['kw_args'][key] = 'Error: ' + str(ex)
+        
+        # get ready to serialize result
+        resdict = {}
+        
+        try:
+            plan_fn = getattr(self.planner, method)
+            traj = plan_fn(*args, **kw_args)
+            resdict['ok'] = True
+            resdict['traj_first'] = list(map(float,traj.GetWaypoint(0)))
+            resdict['traj_last'] = list(map(float,traj.GetWaypoint(traj.GetNumWaypoints()-1)))
+            return traj
+            
+        except prpy.planning.PlanningError as ex:
+            resdict['ok'] = False
+            resdict['exception'] = str(ex)
+            raise
+        
+        except:
+            resdict['ok'] = False
+            resdict['exception'] = 'unknown exception type'
+            raise
+        
+        finally:
+            # create yaml dictionary to be serialized
+            yamldict = {}
+            yamldict['environment'] = envdict
+            yamldict['request'] = reqdict
+            yamldict['result'] = resdict
+            fp = open(fn,'w')
+            yaml.safe_dump(yamldict, fp)
+            fp.close()
+        
+        
--- a/src/prpy/serialization.py
+++ b/src/prpy/serialization.py
@@ -105,7 +105,10 @@ def serialize_kinbody(body, uri_only=False):
 
     uri = body.GetXMLFilename()
     
-    if not uri:
+    if uri_only and not uri:
+        serialization_logger.warn(
+            'uri_only passed, but kinbody {} doesn\'t have a URI.'.format(
+            body.GetName()))
         uri_only = False
     
     data = {
@@ -113,7 +116,8 @@ def serialize_kinbody(body, uri_only=False):
         'name': body.GetName(),
         'uri': uri,
     }
-    
+
+    # only add uri_only to serialization so that older deserializers work
     if uri_only:
         data['uri_only'] = True
     
@@ -149,12 +153,10 @@ def serialize_kinbody_state(body, uri_only=False):
 
     link_transforms, dof_branches = body.GetLinkTransformations(True)
     data.update({
+        'link_transforms': map(serialize_transform, link_transforms),
+        'dof_branches': dof_branches.tolist(),
         'dof_values': body.GetDOFValues().tolist(),
     })
-    if not uri_only:
-        data.update({
-            'link_transforms': map(serialize_transform, link_transforms),
-        })
 
     return data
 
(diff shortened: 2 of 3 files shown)


src/prpy/serialization.py
```diff
@@ -99,43 +101,60 @@ def serialize_environment_file(env, path, writer=None):
...
-def serialize_kinbody(body):
-    all_joints = []
-    all_joints.extend(body.GetJoints())
-    all_joints.extend(body.GetPassiveJoints())
+def serialize_kinbody(body, uri_only=False):
 
+    uri = body.GetXMLFilename()
+    
+    if not uri:
+        uri_only = False
```
mkoval: Print a warning in this case. It's odd that you can pass `uri_only`, but then have the implementation silently fall back on full serialization. Or add another option like `try_uri` that makes it clear that it's best-effort only.

src/prpy/serialization.py
```diff
@@ -99,43 +101,60 @@ def serialize_environment_file(env, path, writer=None):
...
         'name': body.GetName(),
-        'uri': body.GetXMLFilename(),
-        'links': map(serialize_link, body.GetLinks()),
-        'joints': map(serialize_joint, all_joints),
+        'uri': uri,
     }
-    data['kinbody_state'] = serialize_kinbody_state(body)
+    
+    if uri_only:
+        data['uri_only'] = True
```
mkoval: Why not define `uri_only` in both cases, i.e. `data['uri_only'] = uri_only`?

src/prpy/serialization.py
```diff
@@ -99,43 +101,60 @@ def serialize_environment_file(env, path, writer=None):
...
+def serialize_kinbody_state(body, uri_only=False):
     data = {
         name: get_fn(body)
         for name, (get_fn, _) in KINBODY_STATE_MAP.iteritems()
     }
 
     link_transforms, dof_branches = body.GetLinkTransformations(True)
     data.update({
-        'link_transforms': map(serialize_transform, link_transforms),
-        'dof_branches': dof_branches.tolist(),
```
mkoval: Why did you remove `dof_branches`?

src/prpy/serialization.py
```diff
@@ -99,43 +101,60 @@ def serialize_environment_file(env, path, writer=None):
...
         'dof_values': body.GetDOFValues().tolist(),
     })
+    if not uri_only:
+        data.update({
+            'link_transforms': map(serialize_transform, link_transforms),
+        })
```
mkoval: `link_transformations` and `dof_branches` are needed to call `SetDOFValues` and get back bit-identical results. They should be serialized _even if_ `uri_only` is set.

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
-        kinbody = RaveCreateRobot(env, '')
-        kinbody.Init(
-            link_infos, joint_infos,
-            manipulator_infos, sensor_infos,
-            data['uri']
-        )
+    
+    name_desired = name or data['name']
+    
+    if 'uri_only' in data and data['uri_only']:
```
mkoval: Replace this with `data.get('uri_only', False)`.

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
+        
+        if data['is_robot']:
+            
+            if data['uri'].endswith('.robot.xml'):
+                kinbody = env.ReadRobotXMLFile('robots/barrettwam.robot.xml')
```
mkoval: Uhhhh..what?

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
+                kinbody.SetName(name_desired)
+                env.Add(kinbody, anonymous)
+            else:
```
mkoval: I would prefer to flip these two cases. OpenRAVE supports several file extensions (`.dae`, `.zae`, etc) that would not be correctly handled by this. I would prefer to _only_ use `or_urdf` if the file ends in `.urdf` or `.srdf`.

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
+                urdf,srdf = data['uri'].split()
```
mkoval: There is no point in `split()`ing this because you pass it directly to `or_urdf` a few lines below.

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
+                m_urdf = openravepy.RaveCreateModule(env, 'urdf')
```
mkoval: Is there any way we can create _one_ `or_urdf` module for the entire de-serialization, instead of creating a new one for each robot?

src/prpy/serialization.py
```diff
@@ -327,37 +344,66 @@ def deserialize_kinbody(env, data, name=None, anonymous=False, state=True):
...
+                robot_name = m_urdf.SendCommand('Load {:s} {:s}'.format(urdf,srdf))
+                if robot_name != name_desired:
+                    raise RuntimeError('error, or_urdf name mismatch!')
+                kinbody = env.GetRobot(robot_name)
+            
+        else:
+            
+            kinbody = env.ReadKinBodyXMLFile(data['uri'])
+            kinbody.SetName(name_desired)
+            env.Add(kinbody)
```
mkoval: Why don't you handle `or_urdf` in this case?

src/prpy/serialization.py
```diff
@@ -380,11 +426,14 @@ def deserialize_kinbody_state(body, data):
...
-        data['dof_branches']
-    )
+    
+    body.SetDOFValues(data['dof_values'])
+    
+    if 'link_transforms' in data:
+        body.SetLinkTransformations(
+            map(deserialize_transform, data['link_transforms']),
+            data['dof_branches']
+        )
```
mkoval: Please remove `SetDOFValues` and restore `SetLinkTransformations`: `SetDOFValues(GetDOFValues(...))` is not an identity-preserving operation.

gilwoolee:
Thanks @cdellin ! It's been very helpful. As an extensive user to these additional features, I have one suggestion. As you already acknowledged in the PR comment, `replay-log.py`'s robot contains no information other than those from URDF, so for example it can't do `PlanToNamedConfiguration` because the robot does not have config_library. @Shushman was at some point confused about this too. I think it's confusing because it successfully loads the robot and is capable of handling some of the planning queries. :) 

There were a few other features I needed so I ended up just loading HERB from herbpy. I think this script serves its purpose as a base script to work from, but I think having an additional comment at the top of `replay-log.py` explaining that _downstream users may need to add some additional features or load robot based on the retrieved name_ may be useful.

src/prpy/serialization.py
cdellin: Fixed!

src/prpy/serialization.py
cdellin: Added a comment; this is so that old deserializers still work.

src/prpy/serialization.py
cdellin: Fixed! They were removed because I originally wanted to replicate just loading the robot and setting DOF values.

src/prpy/serialization.py
cdellin: Fixed.

src/prpy/serialization.py
cdellin: I only slightly prefer my way, but I'm happy to change `(-:`.

src/prpy/serialization.py
cdellin: Ugh, I *guess* I can remove this.  But seriously, when was the last time that you wanted to load an XML file and it wasn't the WAM?  I think it's actually more likely that you wanted to load the WAM but got the path wrong (which this will fix!) than that you want to load some other robot. `(-:`

src/prpy/serialization.py
cdellin: Fixed.

src/prpy/serialization.py
cdellin: Fixed.

src/prpy/serialization.py
cdellin: Behold the `UnitaryMemoizer` -- it will fix all our problems~

src/prpy/serialization.py
cdellin: Added!

src/prpy/serialization.py
cdellin: I added a case for both, so we can handle the ISER logs.

cdellin:
Thanks @gilwoolee!  I added a comment to the top of `replay-log.py` to that effect.

cdellin:
OK, I think everything's addressed.

src/prpy/planning/logged.py
```diff
@@ -0,0 +1,102 @@
...
+
+import openravepy
+import numpy
+import yaml
+
+import prpy.planning.base
+import prpy.serialization
+
+
+class LoggedPlanner(prpy.planning.base.MetaPlanner):
```
mkoval: Please add a docstring.

src/prpy/planning/logged.py
```diff
@@ -0,0 +1,102 @@
...
+                reqdict['args'].append(prpy.serialization.serialize(arg))
+            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+                reqdict['args'].append(str(ex))
+        reqdict['kw_args'] = {}
+        for key,value in kw_args.items():
+            key = prpy.serialization.serialize(key)
+            try:
+                reqdict['kw_args'][key] = prpy.serialization.serialize(value)
+            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+                reqdict['kw_args'][key] = 'Error: ' + str(ex)
```
mkoval: Why are failures treated differently for `args` and `kwargs` (one is prefixed with `Error: `)? Also, I would prefer to factor this `try`/`except` block into a separate function to remove some copy-pasta.

src/prpy/planning/logged.py
```diff
@@ -0,0 +1,102 @@
...
+        reqdict['method'] = method # string
+        reqdict['args'] = []
+        for arg in args:
+            try:
+                reqdict['args'].append(prpy.serialization.serialize(arg))
+            except prpy.exceptions.UnsupportedTypeSerializationException as ex:
+                reqdict['args'].append(str(ex))
+        reqdict['kw_args'] = {}
+        for key,value in kw_args.items():
+            key = prpy.serialization.serialize(key)
```
mkoval: Missing error checking on the dictionary key.

src/prpy/planning/logged.py
mkoval: You could even replace all of this with a wicked dictionary comprehension if you add that function! :grin:

src/prpy/planning/logged.py
```diff
@@ -0,0 +1,102 @@
...
+        
+        # get ready to serialize result
+        resdict = {}
+        
+        try:
+            plan_fn = getattr(self.planner, method)
+            traj = plan_fn(*args, **kw_args)
+            resdict['ok'] = True
+            resdict['traj_first'] = list(map(float,traj.GetWaypoint(0)))
+            resdict['traj_last'] = list(map(float,traj.GetWaypoint(traj.GetNumWaypoints()-1)))
```
mkoval: Why not serialize the entire trajectory?

src/prpy/serialization.py
```diff
@@ -16,7 +16,9 @@ def serialize(obj):
 
     NoneType = type(None)
 
-    if isinstance(obj, (int, float, basestring, NoneType)):
+    if isinstance(obj, numpy.float64):
```
mkoval: Replace `numpy.float65` with `numpy.floating`, which is the base class for all NumPy floating point types (`float`, `float16`, `float32`, `float64`, and `float128`. You may want to add similar logic to convert `numpy.signedinteger` and `numpy.unsignedinteger` to `int`s.

src/prpy/serialization.py
```diff
@@ -99,32 +101,51 @@ def serialize_environment_file(env, path, writer=None):
...
-    all_joints.extend(body.GetJoints())
-    all_joints.extend(body.GetPassiveJoints())
+def serialize_kinbody(body, uri_only=False):
 
+    uri = body.GetXMLFilename()
+    
+    if uri_only and not uri:
+        serialization_logger.warn(
+            'uri_only passed, but kinbody {} doesn\'t have a URI.'.format(
+            body.GetName()))
```
mkoval: Nit: `kinbody` -> `KinBody`. Also, I would prefer to specifically mention that `GetXMLFilename()` is empty. This will give the user a hint of where to look while debugging this.

src/prpy/serialization.py
```diff
@@ -99,32 +101,51 @@ def serialize_environment_file(env, path, writer=None):
...
         'is_robot': body.IsRobot(),
         'name': body.GetName(),
-        'uri': body.GetXMLFilename(),
-        'links': map(serialize_link, body.GetLinks()),
-        'joints': map(serialize_joint, all_joints),
+        'uri': uri,
     }
-    data['kinbody_state'] = serialize_kinbody_state(body)
+
+    # only add uri_only to serialization so that older deserializers work
```
mkoval: Comment doesn't make sense. I think you mean:

> Only add `uri_only` to the serialization **if `uri_only` is set** so older deserializers work.

src/prpy/serialization.py
```diff
@@ -99,32 +101,51 @@ def serialize_environment_file(env, path, writer=None):
...
         'robot_state': serialize_robot_state(robot),
     }
+    if not uri_only:
+        data.update({
+            'manipulators': map(serialize_manipulator, robot.GetManipulators()),
+        })
+    return data
 
-def serialize_kinbody_state(body):
+def serialize_kinbody_state(body, uri_only=False):
```
mkoval: This function no longer needs the `uri_only` parameter. It doesn't make sense conceptually, either, since the state must be serialized even if `uri_only` is `True`.

src/prpy/serialization.py
```diff
@@ -195,6 +216,17 @@ def serialize_transform(t):
     }
 
 # Deserialization.
+
+class UnitaryMemoizer:
```
mkoval: Missing docstring.

src/prpy/serialization.py
```diff
@@ -300,12 +330,16 @@ def deserialize_environment(data, env=None, purge=False, reuse_bodies=None):
             deserialization_logger.debug('Purging body "%s".', body.GetName())
             env.Remove(body)
 
+    # Create a or_ordf module on demand
+    urdf_module_getter = UnitaryMemoizer(lambda: openravepy.RaveCreateModule(env,'urdf'))
```
mkoval: Throw an exception if creating the `urdf` module fails, i.e. `RaveCreateModule` returns `None`.

src/prpy/serialization.py
```diff
@@ -320,44 +354,83 @@ def deserialize_environment(data, env=None, purge=False, reuse_bodies=None):
...
+    
+    if data.get('uri_only', False):
+        
+        if data['is_robot']:
+
+            parts = data['uri'].split()
+            if len(parts)==2 and parts[0].endswith('.urdf') and parts[1].endswith('.srdf'):
+                robot_name = urdf_module_getter().SendCommand('Load {}'.format(data['uri']))
+                if robot_name != name_desired:
+                    raise RuntimeError('error, or_urdf name mismatch!')
```
mkoval: This check is incorrect. `or_urdf` loads the robot with a default name generated from the URDF file. This code should call `kinbody.SetName(name_desired)` to override the name with the serialized value.

src/prpy/serialization.py
```diff
@@ -320,44 +354,83 @@ def deserialize_environment(data, env=None, purge=False, reuse_bodies=None):
...
+                kinbody = env.ReadRobotXMLFile(data['uri'])
+                kinbody.SetName(name_desired)
+                env.Add(kinbody, anonymous)
+            
+        else:
+
+            if data['uri'].endswith('.urdf'):
+                kinbody_name = urdf_module_getter().SendCommand('Load {}'.format(data['uri']))
+                if kinbody_name != name_desired:
+                    raise RuntimeError('error, or_urdf name mismatch!')
```
mkoval: See above. This logic is incorrect.

src/prpy/serialization.py
```diff
@@ -320,44 +354,83 @@ def deserialize_environment(data, env=None, purge=False, reuse_bodies=None):
...
+                data['uri']
+            )
+        
+        else:
+            
+            kinbody = RaveCreateKinBody(env, '')
+            kinbody.Init(link_infos, joint_infos, data['uri'])
+        
+        kinbody.SetName(name_desired)
+        env.Add(kinbody, anonymous)
```
mkoval: All of this code appears to be mistakenly in the `if uri_only` block.

src/prpy/serialization.py
```diff
@@ -381,10 +454,19 @@ def deserialize_kinbody_state(body, data):
...
-        data['dof_branches']
-    )
+    link_transforms = data.get('link_transforms')
+    dof_branches = data.get('dof_branches')
+    if link_transforms is not None and dof_branches is not None:
+        body.SetLinkTransformations(
+            map(deserialize_transform, link_transforms),
+            dof_branches
+        )
+
```
mkoval: Nit: Remove newline.

src/prpy/serialization.py
```diff
@@ -381,10 +454,19 @@ def deserialize_kinbody_state(body, data):
...
+    else:
+        deserialization_logger.warn(
+            ('KinBody "{}" does not have link_transforms/dof_branches'
+            + 'saved; falling back to dof_values').format(body.GetName()))
```
mkoval: This awkward string concatenation is not necessary. Python does string concatenation like C (your favorite, @cdellin!):
```
>>> print ('banana'
           'orange')
bananaorange

**Comment by mkoval** [comment]
Added another round of comments.

File: src/prpy/planning/logged.py (hunk 12 above)
Comment (cdellin) [inline src/prpy/planning/logged.py]: Added.

File: src/prpy/planning/logged.py (hunk 13 above)
Comment (cdellin) [inline src/prpy/planning/logged.py]: Done! No wicked dictionary comprehension though ... sounds hard to read. (Feel free to try it if you want!)

File: src/prpy/planning/logged.py (hunk 14 above)
Comment (cdellin) [inline src/prpy/planning/logged.py]: Added a comment; this will through a serialization error on failure, which is I think what we want.

File: src/prpy/planning/logged.py (hunk 15 above)
Comment (cdellin) [inline src/prpy/planning/logged.py]: Done!

File: src/prpy/serialization.py (hunk 16 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Done.

File: src/prpy/serialization.py (hunk 17 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Good idea! Done.

File: src/prpy/serialization.py (hunk 18 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Fixed!

File: src/prpy/serialization.py (hunk 19 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Yuppers!

File: src/prpy/serialization.py (hunk 20 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Missing no longer. (Is `UnitaryMemoizer` not sufficiently self-describing? `d-:`)

File: src/prpy/serialization.py (hunk 21 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Done (see later on use).

File: src/prpy/serialization.py (hunk 22 above)
Comment (cdellin) [inline src/prpy/serialization.py]: I think I fixed this ... (I need to remove from environment to change name, no?)

File: src/prpy/serialization.py (hunk 23 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Ditto.

File: src/prpy/serialization.py (hunk 24 above)
Comment (cdellin) [inline src/prpy/serialization.py]: I looked at this, and I think it's correct ... there's an `else:` above, so this code should only be executed if `uri_only` is not set.

File: src/prpy/serialization.py (hunk 25 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Newline obliterated.

File: src/prpy/serialization.py (hunk 26 above)
Comment (cdellin) [inline src/prpy/serialization.py]: Fancy! Done and done.

**Comment by cdellin** [comment]
OK, took care of everything (besides the non-`uri_only` kinbody loading logic which I contend is correct).  I'm ready for round 3!

File: src/prpy/planning/logged.py [hunk 27]
```diff
@@ -0,0 +1,145 @@
...
+        # retrieve enviroment
+        if 'robot' in kw_args:
+            env = kw_args['robot'].GetEnv()
+        elif len(args) > 0:
+            env = args[0].GetEnv()
+        else:
+            raise RuntimeError('could not retrieve env from planning request!')
+        
+        # serialize environment
+        envdict = prpy.serialization.serialize_environment(env, uri_only=True)
```
mkoval: Can you make `uri_only` a constructor parameter so it's not hardcoded to be `True`?

src/prpy/planning/logged.py
```diff
@@ -0,0 +1,145 @@
...
+            return traj
+            
+        except prpy.planning.PlanningError as ex:
+            resdict['ok'] = False
+            resdict['exception'] = str(ex)
+            raise
+        
+        except:
+            resdict['ok'] = False
+            resdict['exception'] = 'unknown exception type'
```
mkoval: You can get the exception here using `str(sys.exc_info[1])` and eliminate the `PlanningError` case.

src/prpy/serialization.py
```diff
@@ -99,30 +103,51 @@ def serialize_environment_file(env, path, writer=None):
...
-    all_joints.extend(body.GetJoints())
-    all_joints.extend(body.GetPassiveJoints())
+def serialize_kinbody(body, uri_only=False):
 
+    uri = body.GetXMLFilename()
+    
+    if uri_only and not uri:
+        serialization_logger.warn(
+            'uri_only passed, but KinBody "{}"\'s GetXMLFilename()'
+            'returns an empty URI.'.format(
```
mkoval: Missing space before `returns`.

src/prpy/serialization.py
```diff
@@ -381,10 +475,18 @@ def deserialize_kinbody_state(body, data):
...
+    dof_branches = data.get('dof_branches')
+    if link_transforms is not None and dof_branches is not None:
+        body.SetLinkTransformations(
+            map(deserialize_transform, link_transforms),
+            dof_branches
+        )
+    else:
+        deserialization_logger.warn(
+            'KinBody "{}" does not have link_transforms/dof_branches'
+            'saved; falling back to dof_values'.format(body.GetName()))
```
mkoval: Missing a space before `saved`.

src/prpy/planning/logged.py
cdellin: Added!

src/prpy/planning/logged.py
cdellin: I did the former, but I prefer if non-expected exceptions have the additional "Unknown exception type" text.

src/prpy/serialization.py
cdellin: Done!

src/prpy/serialization.py
cdellin: Done!

cdellin:
Comments addressed!

PR #322, opened 2016-07-23
Lazier TSRPlanner refactor

mkoval:
This supersedes #321 by making both IK generation **and** collision checking lazy.

The algorithm now looks like this:

1. Generate a set of `num_candidates` IK solutions without collision checking.
2. Rank the set of candidates by a user-provided score function.
3. Iteratively collision check the candidates in rank order until `chunk_size` are collision free.
4. If necessary, repeat (1)-(3) until the chunk is full or we reach a user-defined time limit.
5. Attempt to `PlanToConfigurations` to the chunk.
6. Repeat (1)-(5) `num_attempts` times.

Note that the sampling timeout is shared _across all attempts_. We may try fewer than `num_attempts` planning queries if we are unable to generate collision free samples quickly enough.

PR #331, opened 2016-07-25
Abstracted robot collision checking into class for snap and vectorfield

cdellin:
In the push for baked checking in PrPy, this is a first pass which creates a two robot collision checkers:
- `SimpleRobotCollisionChecker`
- `BakedRobotCollisionChecker`
Each of these can be constructed for a robot, and expose a simple `CheckCollision` method, which raises a PrPy exception upon collision.

This PR also refactors the `SnapPlanner` and `VectorFieldPlanner` planners to make use of this abstraction.  The planners default to `SimpleRobotCollisionChecker`, but a different type can be passed in when they are instantiated, and this type will be used for its collision checking.

Note that the `BakedRobotCollisionChecker` is experimental, and only works with the `baked_checkers` branch of `or_fcl` at the moment.  Currently, the underlying checker does not expose which links provoked the collision, so the collision report is not helpful, and a vanilla `CollisionPlanningError` is currently thrown.

I'll wait to implement this for `TSRPlanner` until its PR (#322) is merged.

### DIFF
--- a/src/prpy/collision.py
+++ b/src/prpy/collision.py
@@ -98,9 +98,7 @@ class BakedRobotCollisionChecker:
 
     def CheckCollision(self, report=None):
         # The baked check is performed by checking self collision on baked
-        if self.checker.CheckSelfCollision(self.baked, report):
-            return True
-        return False
+        return self.checker.CheckSelfCollision(self.baked, report):
 
     def VerifyCollisionFree(self):
         report = openravepy.CollisionReport()
--- a/src/prpy/collision.py
+++ b/src/prpy/collision.py
@@ -1,0 +1,65 @@
+# Copyright (c) 2013, Carnegie Mellon University
+# All rights reserved.
+# Authors: Chris Dellin <cdellin@gmail.com>
+#          Michael Koval <mkoval@cs.cmu.edu>
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+# - Redistributions of source code must retain the above copyright notice, this
+#   list of conditions and the following disclaimer.
+# - Redistributions in binary form must reproduce the above copyright notice,
+#   this list of conditions and the following disclaimer in the documentation
+#   and/or other materials provided with the distribution.
+# - Neither the name of Carnegie Mellon University nor the names of its
+#   contributors may be used to endorse or promote products derived from this
+#   software without specific prior written permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import openravepy
+from prpy.planning.exceptions import CollisionPlanningError
+from prpy.planning.exceptions import SelfCollisionPlanningError
+
+
+class SimpleRobotCollisionChecker:
+
+    def __init__(self, robot):
+        self.robot = robot
+        self.env = robot.GetEnv()
+
+    def CheckCollision(self):
+        report = openravepy.CollisionReport()
+        if self.env.CheckCollision(self.robot, report=report):
+            raise CollisionPlanningError.FromReport(report)
+        elif self.robot.CheckSelfCollision(report=report):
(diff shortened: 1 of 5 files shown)


src/prpy/collision.py
```diff
@@ -0,0 +1,65 @@
...
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import openravepy
+from prpy.planning.exceptions import CollisionPlanningError
+from prpy.planning.exceptions import SelfCollisionPlanningError
+
+
+class SimpleRobotCollisionChecker:
```
mkoval: Missing docstring.

src/prpy/collision.py
```diff
@@ -0,0 +1,65 @@
...
+        elif self.robot.CheckSelfCollision(report=report):
+            raise SelfCollisionPlanningError.FromReport(report)
+
+
+class BakedRobotCollisionChecker:
+
+    def __init__(self, robot):
+        self.robot = robot
+        self.env = robot.GetEnv()
+        self.checker = self.env.GetCollisionChecker()
```
mkoval: Can you check the type of `GetCollisionChecker` before trying to use it?

src/prpy/collision.py
```diff
@@ -0,0 +1,65 @@
...
+        kb_type = self.checker.SendCommand('BakeGetType')
+        self.checker.SendCommand('BakeBegin')
+        self.env.CheckCollision(self.robot)
+        self.robot.CheckSelfCollision()
+        self.baked = openravepy.RaveCreateKinBody(self.env, kb_type)
```
mkoval: Check for an error here.

src/prpy/collision.py
```diff
@@ -0,0 +1,65 @@
...
+        self.checker.SendCommand('BakeEnd')
+
+    def CheckCollision(self):
+        report = openravepy.CollisionReport()
+        if self.checker.CheckSelfCollision(self.baked, report):
```
mkoval: Please add a comment explaining that this performs the baked check.

src/prpy/collision.py
```diff
@@ -0,0 +1,65 @@
...
+class BakedRobotCollisionChecker:
+
+    def __init__(self, robot):
+        self.robot = robot
+        self.env = robot.GetEnv()
+        self.checker = self.env.GetCollisionChecker()
+        kb_type = self.checker.SendCommand('BakeGetType')
+        self.checker.SendCommand('BakeBegin')
+        self.env.CheckCollision(self.robot)
+        self.robot.CheckSelfCollision()
```
mkoval: Please add a comment explaining that this bakes the two collision checks.

src/prpy/planning/snap.py
```diff
@@ -36,6 +36,7 @@
     ClonedPlanningMethod,
     Tags
 )
+from prpy.collision import SimpleRobotCollisionChecker
```
mkoval: Prefer: `from ..collision import SimpleRobotCollisionChecker`

src/prpy/planning/vectorfield.py
```diff
@@ -35,6 +35,7 @@
 import openravepy
 from .. import util
 from base import BasePlanner, PlanningError, ClonedPlanningMethod, Tags
+from prpy.collision import SimpleRobotCollisionChecker
```
mkoval: Prefer: `from ..collision import SimpleRobotCollisionChecker`

mkoval:
Looks good, other than the minor comments I left above. Can you also add this to CHOMP, GreedyIk, and TSRPlanner?

src/prpy/collision.py
cdellin: Added docstrings.

src/prpy/collision.py
cdellin: I don't check the type, but I do now check for the case where no checker exists, and I raise a more helpful exception when the checker doesn't support baked checking.

src/prpy/collision.py
cdellin: Once we've confirmed that the checker supports baking, this is assumed to work.  If not, it already throws an exception (kinbody type not known!) which is correct.

src/prpy/collision.py
cdellin: Comment added!

src/prpy/collision.py
cdellin: Comment added.

src/prpy/planning/snap.py
cdellin: Fixed (against my better judgment `(-:`).

cdellin:
I addressed most of the comments, and also added this for `ChompPlanner`.

I'd prefer to wait to add this to `TSRPlanner` until #322 is merged.  Also, the current implementation of `GreedyIKPlanner` does its collision checking in `FindIKSolutions`, and unfortunately there's no way to add baked support to that planner.

cdellin:
I like how immediately after my last comment, you went ahead and merged #322.  Fine!  `TSRPlanner` support is now in.  (I refactored the robot collision interface to make it a bit nicer and more consistent with the native OpenRAVE interface.)

src/prpy/collision.py
```diff
@@ -0,0 +1,108 @@
...
+        self.baked = openravepy.RaveCreateKinBody(self.env, kb_type)
+        self.checker.SendCommand('BakeEnd')
+
+    def CheckCollision(self, report=None):
+        # The baked check is performed by checking self collision on baked
+        if self.checker.CheckSelfCollision(self.baked, report):
+            return True
+        return False
```
mkoval: Replace this with:
```python
return self.checker.CheckSelfCollision(self.baked, report)
```
:grin:

src/prpy/planning/tsr.py
```diff
@@ -136,12 +138,16 @@ def compute_ik_solutions(tsrchain):
 
             return ik_solutions
 
+        # Instantiate a robot checker
+        with CollisionOptionsStateSaver(
+                self.env.GetCollisionChecker(), CollisionOptions.ActiveDOFs):
+            robot_checker = self.robot_collision_checker(robot)
+
```
mkoval: You should move the `SetActiveDOFs` call above the baking.

src/prpy/collision.py
cdellin: Haha done!

src/prpy/planning/tsr.py
cdellin: Good catch! Fixed.

cdellin:
Comments addressed!

PR #332, opened 2016-07-26
Switched to L2 DOF resolution in VectorField and GreedyIk

jeking04:
This PR updates Vectorfield and GreedyIK to use L2 norm for determining collision checking resolution.

### DIFF
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -528,11 +528,14 @@ class VectorFieldPlanner(BasePlanner):
                 else:
                     # TODO: This will recheck the entire trajectory
                     #  Ideally should just check the new portion of the trajectory
-                    from prpy.util import VanDerCorputSampleGenerator
-                    vdc = VanDerCorputSampleGenerator
+
+                    # Use SampleTimeGenerator so that the time points we get
+                    # to check are in order
+                    from prpy.util import SampleTimeGenerator
+                    stg = SampleTimeGenerator
                     checks = GetLinearCollisionCheckPts(robot, path,
                                                         norm_order=2,
-                                                        sampling_func=vdc)
+                                                        sampling_func=stg)
                     # start_time=nonlocals['t_check'])
 
                 for t_check, q_check in checks:
--- a/src/prpy/planning/vectorfield.py
+++ b/src/prpy/planning/vectorfield.py
@@ -439,7 +439,7 @@ class VectorFieldPlanner(BasePlanner):
             SelfCollisionPlanningError,
         )
         from openravepy import CollisionReport, RaveCreateTrajectory
-        from ..util import GetCollisionCheckPts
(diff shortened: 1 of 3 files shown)


src/prpy/planning/vectorfield.py
```diff
@@ -526,11 +526,13 @@ def fn_callback(t, q):
...
-                    # loop.
-                    checks = GetCollisionCheckPts(robot, path,
-                                                  include_start=False)
+                    # TODO: This will recheck the entire trajectory
+                    #  Ideally should just check the new portion of the trajectory
+                    from prpy.util import VanDerCorputSampleGenerator
+                    vdc = VanDerCorputSampleGenerator
+                    checks = GetLinearCollisionCheckPts(robot, path,
+                                                        norm_order=2,
+                                                        sampling_func=vdc)
```
mkoval: Can you modify `FollowVectorField` to accept a `sampling_func`? The default can still be `VanDerCorputSampleGenerator`. This is the convention we've followed in other planners to allow the user to configure this option.

src/prpy/planning/workspace.py
```diff
@@ -206,11 +213,14 @@ def PlanWorkspacePath(self, robot, traj, timelimit=5.0,
                     infeasible_step = True
                     if qnew is not None:
                         # Found an IK
-                        step = abs(qnew - qcurr)
-                        if (max(step) < min_step) and qtraj:
+                        steps = abs(qnew - qcurr) / q_resolutions;
+                        norm = numpy.linalg.norm(steps, ord=norm_order)
+
+                        if (norm < min_step) and qtraj:
```
mkoval: I think this dramatically changed the scaling on `min_step` since the norm is now scaled by DOF resolution. Can you adjust the value of that constant accordingly?

src/prpy/planning/vectorfield.py
```diff
@@ -526,11 +528,10 @@ def fn_callback(t, q):
...
                 else:
-                    # TODO: This should start at t_check. Unfortunately, a bug
-                    # in GetCollisionCheckPts causes this to enter an infinite
-                    # loop.
-                    checks = GetCollisionCheckPts(robot, path,
-                                                  include_start=False)
+                    # TODO: This will recheck the entire trajectory
+                    #  Ideally should just check the new portion of the trajectory
+                    checks = GetLinearCollisionCheckPts(robot, path,
+                            norm_order=norm_order, sampling_func=sampling_func)
```
mkoval: Unfortunately, this isn't enough. `GetLinearCollisionCheckPts` retimes the trajectory internally and returns timestamps on the retimed trajectory. This means that the value of `t_check` below does not correspond to the original trajectory.

This change causes `VectorFieldPlanner` to fail a unit test about constraint deviation because the trajectory is being sampled at arbitrary (incorrect) times.

PR #333, opened 2016-07-26
Enable CBiRRT L-2 norm for dof resolution

cdellin:
A recent [comps PR](https://github.com/personalrobotics/comps/pull/7) adds a flag (off by default) to switch CBiRRT's interpretation of the robot's DOF resolutions from an L-infinity norm to an L-2 norm.  Once that's merged, we should merge this to turn it on from PrPy (mirroring #332 for the Python planners).

### DIFF
--- a/src/prpy/planning/cbirrt.py
+++ b/src/prpy/planning/cbirrt.py
@@ -198,6 +198,10 @@ class CBiRRTPlanner(BasePlanner):
 
         args = ['RunCBiRRT']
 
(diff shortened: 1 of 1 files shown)


cdellin:
I added the `steplength` parameter to CBiRRT. I chose `0.0599` (was `0.05` before by default) since it was the closest to the original that is just under a multiple of Herb's DOF resolution (`0.02`).

PR #342, opened 2016-08-01
[WIP] RobotCollisionChecker refactor

mkoval:


src/prpy/collision.py
```diff
@@ -78,29 +103,95 @@ class BakedRobotCollisionChecker:
...
+
         self.checker = self.env.GetCollisionChecker()
         if self.checker is None:
             raise PrPyException('No collision checker found on environment')
+
+        self.baked_kinbody = None
+        self.collision_saver = CollisionOptionsStateSaver(
+            self.checker, collision_options)
+
+    def __enter__(self):
```
cdellin: Are these duplicated `__enter__` and `__exit__` methods intentional?

cdellin:
Except the one issue above, I like this change.  I have three comments:
- The naming is a bit weird; the `CollisionChecker` objects which are passed as arguments named `robot_collision_checker` are actually factories which produce context managers, but when instantiated in `with self.robot_collision_checker(robot) as robot_checker:`, the context manager is called a checker.
- I don't immediately know why it's a good idea to pull the `CollisionOptionsStateSaver` into the robot state checker.  Why not keep the robot checker collision option agnostic, and use nested managers to handle that? If it's just for convenience, maybe you can just define nested versions which just call `nested()`?
- The wrapper (factory) classes seem a bit strange ... why not just pass the context manager class name as an argument? Is it so you can specify a default value for `collision_options`? Oh, perhaps it's so you can extract the desired collision options in case the planner wants to handle them differently?

cdellin:
Ok, I thought about this a bit, and this seems like the most reasonable solution (although the naming issue in my first bullet is still a bit confusing).

src/prpy/collision.py
mkoval: Fixed. Good catch. 👍

mkoval:
> The naming is a bit weird [...]

I agree that the naming is a mess. It was already confusing and introducing an extra level of factories makes it more so. Any suggestions about how we can rename the factory and the context manager to make this more clear?

> Why not keep the robot checker collision option agnostic, and use nested managers to handle that?

This would require that we create a `nested` context manager factory. It would also mean that we can't easily use the same API on C++ planners, like `OMPL` and `CBiRRT`, using `isinstance` checks.

> Is it so you can specify a default value for `collision_options`?

Yes. In the future, we might want to do more here. For example, we could configure how the baking is done.

PR #345, opened 2016-08-09
Renamed collision checker context managers

mkoval:
This addresses the naming issues introduced by #342. There is now a `xRobotCollisionCheckerFactory` that creates a `xRobotCollisionChecker`. We now use the names `robot_checker_factory` and `robot_checker`, respectively, to refer to instances of these classes.


ClintLiddick:
:+1:

cdellin:
LGTM!

PR #346, opened 2016-08-11
Don't depend on filename to determine robot name in tsrlibrary

ClintLiddick:
Robots all implement `GetName()`, so we should use the names directly instead of inferring from a filename. I changed this to support models loaded via new personalrobotics/or_urdf method `LoadJsonString` https://github.com/personalrobotics/or_urdf/pull/26

@jeking04 @mkoval is there a reason not to do this?

### DIFF
--- a/src/prpy/tsr/tsrlibrary.py
+++ b/src/prpy/tsr/tsrlibrary.py
@@ -69,7 +69,7 @@ class TSRLibrary(object):
         if robot_name is not None:
             self.robot_name = robot_name
         else:
-            self.robot_name = self.get_object_type(robot)
+            self.robot_name = robot.GetName()
             logger.debug('Inferred robot name "%s" for TSRLibrary.', self.robot_name)
 
     def clone(self, cloned_robot):


jeking04:
Yes, there is a reason not to do this. The name of the robot can be changed arbitrarily through a call to `SetName`. If the caller does this, then there will be no TSRs registered for the robot unless the programmer manually re-registers all TSRs against this name.  By using the underlying file that is loaded, we are free to set names of robots as we like and still be able to easily get the correct TSRs.

One feasible reason we would need to change the name of the robot was if we wanted to load more than one into an environment, i.e. we wanted two HERBs to interact with each other.  OpenRAVE will not allow you to have to kinbodies with the same name.

ClintLiddick:
Do you know of another piece of data that allows us to get the name but avoid parsing the filename? In order to support multiple herb models we need to load the robot description from the parameter server, so we literally don't have a filename to parse. And if we did, the filename might end up being `herb_no_left_bh280_right.urdf`. 
I don't know how often we load multiple identical robots, but perhaps a naming convention + search or regex in the tsrlibrary would suffice? I.e. `herb_1` and `herb_2` could be parsed with `^([a-zA-Z0-9]+)_.*` to extract the robot name.

ClintLiddick:
Closing in favor of passing in name from herbpy

mkoval:
I second @jeking04's concern. `TSRLibrary` only uses the file name as a default if `robot_name` is not explicitly set. I would prefer to modify HerbPy to explicitly pass `robot_name='herb'` and modify `TSRLibrary` to throw an exception if it cannot infer a good default.

There are potentially deeper issues lurking in `TSRLibrary`. Knowing that a robot is "HERB" is not sufficient to define a TSR if we do not know what hand(s) are installed. I think @jeking04 has already made some progress towards this.

Issue #341, opened 2016-08-01
robot.PostProcessPath overrides timelimit to default_timelimit=0.5


gilwoolee:
robot.PostProcess sets timelimit to `default_timelimit=0.5`. Unless we pass in `shortcut_options`, `smoothing_options`, `retimer_options` with timelimit, robot.smoother and robot.retimer's timelimit option gets overridden.

mkoval:
Fixed in #349.

PR #347, opened 2016-08-12
Switched to calling robot.retimer on SMOOTH paths

mkoval:
This pull request:

1. Switches to calling `robot.retimer` on `SMOOTH` trajectories. This avoids shortcutting trajectories that are already smooth.
2. Deprecates the `default_timelimit` parameter to `PostProcessPath`, which would insidiously mask the timelimit passed into the post-processor at construction time. See #341.

These changes are backwards compatible. However, you may see poor quality `SMOOTH` trajectory execution from (1) if `robot.retimer` is set to a post-processing algorithm that stops at every waypoint.

### DIFF
--- a/src/prpy/base/robot.py
+++ b/src/prpy/base/robot.py
@@ -219,7 +219,7 @@ class Robot(openravepy.Robot):
         return active_manipulators
 
     def PostProcessPath(self, path,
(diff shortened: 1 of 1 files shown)


jeking04:
:+1:

PR #349, opened 2016-08-15
Added timelimit to CBiRRT constructor

mkoval:
Previously it was only possible to set `timelimit` by passing it to an individual planning call.


jeking04:
LGTM. 👍

PR #351, opened 2016-08-19
Feature/ik planner find iksolutions once

sjavdani:
-Removed SnapPlanner's PlanToEndEffectorPose, which didn't do any ik ranking
-Cleaned up IKPlanner to be in line with other planners
-Also added a new IK ranker that is useful if you know the next configuration you will plan to

Tested with a bunch of planning problems on jaco

### DIFF
--- a/src/prpy/planning/snap.py
+++ b/src/prpy/planning/snap.py
@@ -76,58 +76,6 @@ class SnapPlanner(BasePlanner):
         """
         return self._Snap(robot, goal, **kw_args)
 
-    @ClonedPlanningMethod
-    def PlanToEndEffectorPose(self, robot, goal_pose, **kw_args):
-        """
-        Attempt to plan a straight line trajectory from the robot's
-        current configuration to a desired end-effector pose. This
-        happens by finding the closest IK solution to the robot's
-        current configuration and attempts to snap there
-        (using PlanToConfiguration) if possible.
-        In the case of a redundant manipulator, no attempt is
-        made to check other IK solutions.
-
-        @param robot
-        @param goal_pose desired end-effector pose
-        @return traj
-        """
-        from prpy.planning.exceptions import CollisionPlanningError
-        from prpy.planning.exceptions import SelfCollisionPlanningError
-
-        ikp = openravepy.IkParameterizationType
-        ikfo = openravepy.IkFilterOptions
-
-        # Find an IK solution. OpenRAVE tries to return a solution that is
-        # close to the configuration of the arm, so we don't need to do any
-        # custom IK ranking.
-        manipulator = robot.GetActiveManipulator()
-        ik_param = openravepy.IkParameterization(goal_pose, ikp.Transform6D)
-        ik_solution = manipulator.FindIKSolution(
-            ik_param, ikfo.CheckEnvCollisions,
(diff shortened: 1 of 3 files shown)


mkoval:
LGTM. Thanks @sjavdani!

sjavdani:
Ping! Just making sure this wasn't forgotten about. Can we merge this into master and close?

src/prpy/planning/snap.py
```diff
@@ -76,58 +76,6 @@ def PlanToConfiguration(self, robot, goal, **kw_args):
         """
         return self._Snap(robot, goal, **kw_args)
 
-    @ClonedPlanningMethod
-    def PlanToEndEffectorPose(self, robot, goal_pose, **kw_args):
```
mkoval: I am in favor of eventually removing `PlanToEndEffectorPose` from `SnapPlanner`, but I am concerned about breaking backwards compatibility in the process. 

The simplest solution is to implement a `PlanToEndEffectorPose` wrapper function on `SnapPlanner` that simply calls `_PlanToIK` on `IKPlanner`. In the process, you may want to rename `_PlanToIK` and (possibly) make it a `@classmethod`.

This is how I maintained backwards compatibility for CBiRRT's `PlanToEndEffectorOffset` method after introducing the `PlanToEndEffectorOffsetTSRAdapter` class in #311.

mkoval:
The change overall looks good. On second thought, from my experience in #311, I have one new concern about backwards compatibility. @sjavdani can you address my comment above?

psigen:
Why would this affect backwards compatibility if the previous
implementation did nothing?

mkoval:
The previous implementation did *not* do nothing. However, it *was* missing some key features that @sjavdani needed on ADA (e.g. support for continuous joints, IK ranking). Thanks to @sjavdani, those features are now implemented in `IKPlanner`.

I am in favor of _eventually_ removing `PlanToEndEffectorPose` from `SnapPlanner` and encouraging people to switch to `IKPlanner(delegate_planner=SnapPlanner())`. But doing so will break backwards compatibility unless we take the steps I described above.

psigen:
Is this a hypothetical break, or is there actually some code this affects?

I''d rather not add code bloat forever to support a deprecated use case no
one actually has or in lieu of a one line update in a few of our own
libraries.

Can we forward the method in SnapPlanner to an exception that details how
to update to Shervin's new method?

mkoval:
It looks like @psigen is correct - I can't find a single instance of us using `PlanToEndEffectorPose` on `SnapPlanner` on HERB. We almost exclusively use `TSRLibrary` at this point.

@sjavdani is already switching to this code on ADA, so I retract my comment. We're better off removing the function, i.e. keeping the pull request as it is.

psigen:
Thanks for taking it under consideration @mkoval.

mkoval:
I will merge this once Travis is happy. An unrelated test failed, which I suspect was due to a transient load issue on the VM.

mkoval:
Thanks @sjavdani. Sorry this fell through the cracks for so long.

sjavdani:
No worries. As always, thanks for the careful and thoughtful consideration
that keeps us improving instead of bouncing and breaking things =)

PR #352, opened 2016-08-20
Feature/perception remove all

gilwoolee:
This is a helper method to remove all objects (except `robot` and `kept_bodies`) from environment.

### DIFF
--- a/src/prpy/perception/perception_helper.py
+++ b/src/prpy/perception/perception_helper.py
@@ -28,16 +28,19 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 # POSSIBILITY OF SUCH DAMAGE.
 
-def RemoveAllObjects(env, robot, kept_bodies=[]):
+def RemoveAllObjects(env, robot, kept_bodies=None):
     """
-    Remove everything from the environment except robot and kept_bodies
-    @param env OpenRAVE environment to remove objects
+    Remove everything from the environment except robot and kept_bodies.
+    The environment must be locked while calling this method.
+    @param env OpenRAVE environment to remove objects 
     @param robot Robot in env to keep 
     @param kept_bodies Bodies to keep 
     """
-    with env:
-        for o in env.GetBodies():
-            if o != robot and o not in kept_bodies:
-                env.Remove(o)
+    if kept_bodies is None:
+        kept_bodies = []
+
+    for o in env.GetBodies():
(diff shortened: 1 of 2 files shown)


src/prpy/perception/perception_helper.py
```diff
@@ -0,0 +1,43 @@
...
+# POSSIBILITY OF SUCH DAMAGE.
+
+def RemoveAllObjects(env, robot, kept_bodies=[]):
+    """
+    Remove everything from the environment except robot and kept_bodies
+    @param env OpenRAVE environment to remove objects
+    @param robot Robot in env to keep 
+    @param kept_bodies Bodies to keep 
+    """
+    with env:
```
mkoval: I would prefer to remove the lock from this function and leave that responsibility on the user. This better mirrors the other helper functions, e.g. in `prpy.util`.

src/prpy/perception/perception_helper.py
```diff
@@ -0,0 +1,43 @@
...
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+def RemoveAllObjects(env, robot, kept_bodies=[]):
```
mkoval: It is unsafe to [use a mutable value (e.g. `[]`) as a default value](http://stackoverflow.com/q/1132941/111426) in Python. Please replace this with `kept_bodies=None` and add an `is None` check in the function body to create an empty list.

psigen: Couldn't you just use an immutable empty tuple instead?

mkoval: That's also fine.

PR #353, opened 2016-08-29
Added OMPLRangedPlanner class

### LINKED ISSUE #350: Update RRTConnect wrapper to support a robot_collision_checker

mkoval:
This class adds logic for setting a 'range' parameter such that an
integer number of state validity checks will occur per extension. It
improves on the previous RRTConnect class in two ways:

1. Accepts an optional 'algorithm' parameter, e.g. for PR_RRTConnect.
2. Allows for more than one state validity checks per extension.

This also fixes #350.

### DIFF
--- a/src/prpy/planning/ompl.py
+++ b/src/prpy/planning/ompl.py
@@ -172,21 +172,34 @@ class OMPLRangedPlanner(OMPLPlanner):
 
     This class wraps OMPLPlanner by setting the default 'range' parameter,
     which defines the length of an extension in algorithms like RRT-Connect, to
-    include an fixed number 'multipler' of state validity checks.
+    include an fixed number of state validity checks.
+    
+    This number may be specified explicitly (using 'multiplier') or as an
+    approximate fraction of the longest extent of the state space (using
+    'fraction'). Exactly one of 'multiplier' or 'fraction' is required.
 
     @param algorithm name of the OMPL planner to create
     @param robot_checker_factory robot collision checker factory
     @param multiplier number of state validity checks per extension
+    @param fraction approximate fraction of the maximum extent of the state
+                    space per extension
     """
-    def __init__(self, algorithm='RRTConnect', robot_checker_factory=None,
-            multiplier=1):
-        if not (multiplier >= 1):
-            raise ValueError('Multiplier must be a positive integer.')
+    def __init__(self, multiplier=None, fraction=None,
+                 algorithm='RRTConnect', robot_checker_factory=None):
+        if multiplier is None and fraction is None:
+            raise ValueError('Either "multiplier" of "fraction" is required.')
+        if multiplier is not None and fraction is not None:
+            raise ValueError('"multiplier" and "fraction" are exclusive.')
+        if multiplier is not None and not (multiplier >= 1):
+            raise ValueError('"mutiplier" must be a positive integer.')
+        if fraction is not None and not (0. <= fraction <= 1.):
+            raise ValueError('"fraction" must be between zero and one.')
 
         OMPLPlanner.__init__(self, algorithm='RRTConnect',
                 robot_checker_factory=robot_checker_factory)
+
         self.multiplier = multiplier
-        self.epsilon = 1e-6
+        self.fraction = fraction
 
 
     """ Set a default 'range' parameter from DOF resolutions.
@@ -202,6 +215,8 @@ class OMPLRangedPlanner(OMPLPlanner):
     @return copy of ompl_args with 'range' set to the default value
     """
     def _SetPlannerRange(self, robot, ompl_args=None):
+        epsilon = 1e-6
+
         if ompl_args is None:
             ompl_args = dict()
         else:
@@ -223,15 +238,22 @@ class OMPLRangedPlanner(OMPLPlanner):
                                        - dof_limit_lower[index])
 
             # Duplicate the logic in or_ompl to convert the DOF resolutions to
-            # a fraction of the longest extent of the state space. Then, scale
-            # this by the user-supplied multiple. Finally, subtract a small
-            # value to cope with numerical precision issues inside OMPL.
+            # a fraction of the longest extent of the state space.
             maximum_extent = numpy.max(dof_ranges)
             dof_resolution = robot.GetActiveDOFResolutions()
             conservative_resolution = numpy.min(dof_resolution)
             conservative_fraction = conservative_resolution / maximum_extent
-            ompl_args['range'] = (
-                    self.multiplier * conservative_fraction - self.epsilon)
+
+            if self.multiplier is not None:
+                multiplier = self.multiplier
+            elif self.fraction is not None:
+                multiplier = int(self.fraction / conservative_fraction + 0.5)
+            else:
+                raise ValueError('Either multiplier or fraction must be set.')
+
+            # Then, scale this by the user-supplied multiple. Finally, subtract
+            # a small value to cope with numerical precision issues inside OMPL.
+            ompl_args['range'] = multiplier * conservative_fraction - epsilon
 
         return ompl_args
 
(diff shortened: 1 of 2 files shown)


src/prpy/planning/ompl.py
```diff
@@ -163,10 +164,40 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
+    which defines the length of an extension in algorithms like RRT-Connect, to
+    include an integer number 'multipler' of state validity checks.
+
+    @param algorithm name of the OMPL planner to create
+    @param robot_checker_factory robot collision checker factory
+    @param multiplier number of state validity checks per extension
+    """
+    def __init__(self, algorithm='RRTConnect', robot_checker_factory=None,
+            multiplier=1):
+        if multiplier < 1 or not isinstance(multiplier, int):
```
cdellin: It's a bit weird to order these checks this way; I would reverse it for clarity.

src/prpy/planning/ompl.py
```diff
@@ -163,10 +164,40 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
+                robot_checker_factory=robot_checker_factory)
+        self.multiplier = multiplier
+        self.epsilon = 1e-6
+
+
+    """ Set a default 'range' parameter from DOF resolutions.
 
+    This function returns a copy of the input 'ompl_args' argument with the
+    'range' parameter set to a default value, if it was not already set. The
+    default value is calculated from the active DOFs of 'robot such that there
```
cdellin: Quotation mark syntax error.

src/prpy/planning/ompl.py
```diff
@@ -192,38 +221,34 @@ def _SetPlannerRange(self, robot, ompl_args=None):
             # RRT-Connect's range to the same value used for collision
             # detection inside OpenRAVE.
             longest_extent = numpy.max(dof_ranges)
-            ompl_range = conservative_ratio * longest_extent - epsilon
+            ompl_range  = conservative_ratio * longest_extent - self.epsilon
+            ompl_range *= self.multiplier
```
cdellin: The way I had originally envisioned it was `range = (space_resolution * multipler) - epsilon`, but this should work too.

src/prpy/planning/ompl.py
```diff
@@ -163,10 +164,40 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
+    This class wraps OMPLPlanner by setting the default 'range' parameter,
+    which defines the length of an extension in algorithms like RRT-Connect, to
+    include an integer number 'multipler' of state validity checks.
+
+    @param algorithm name of the OMPL planner to create
+    @param robot_checker_factory robot collision checker factory
+    @param multiplier number of state validity checks per extension
+    """
+    def __init__(self, algorithm='RRTConnect', robot_checker_factory=None,
+            multiplier=1):
```
cdellin: I'm actually a bit wary of the default being 1 here, since that doesn't tend to give good performance.  (Did @gilwoolee do experiments on this for ISER?)  The OMPL default is 20% of the space extent.  Perhaps we could default to something similar?  I'm not sure what the right thing to do is here, but I would hate to see performance suffer as a result of this default.  Or maybe the old code was also effectively defaulting to a unit multiplier?

src/prpy/planning/ompl.py
cdellin: Also, I think the way this code in PrPy calculates the space resolution is inconsistent with the way it's calculated in `or_ompl` (see https://github.com/personalrobotics/or_ompl/blob/master/src/OMPLConversions.cpp#L154). Perhaps we changed the way it's calculated in `or_ompl` at some point without changing it here?

src/prpy/planning/ompl.py
```diff
@@ -192,38 +221,34 @@ def _SetPlannerRange(self, robot, ompl_args=None):
             # RRT-Connect's range to the same value used for collision
             # detection inside OpenRAVE.
             longest_extent = numpy.max(dof_ranges)
-            ompl_range = conservative_ratio * longest_extent - epsilon
+            ompl_range  = conservative_ratio * longest_extent - self.epsilon
+            ompl_range *= self.multiplier
 
             ompl_args['range'] = ompl_range
```
cdellin: Also, I think this `range` parameter is interpreted as a fraction, not as a resolution (in radians).

src/prpy/planning/ompl.py
psigen: Also, doesn't this mean that a `__lt__` or `__cmp__` override on multiplier could actually succeed here even if it is not an `int`?

cdellin: Yeah that's what I was thinking (though the second case would still catch those, it's still a bit confusing).

psigen: Would it? I thought it would short-circuit the `or`...

psigen: Actually, now I'm wondering if we should just not do the check.  If someone wants to pass a dynamically evaluated property or something, I don't see a great reason to shut them down here if it duck-types properly.

cdellin: Maybe replace it with `if not multiplier >= 1:` to allow such types, still catch the typical non-positive case (which might silently yield borked results), and correctly error if `NaN` is passed?

mkoval:
I think I addressed all your concerns. @cdellin and @psigen please review the changes.

src/prpy/planning/ompl.py
```diff
@@ -163,67 +167,115 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
+
+                if joint.IsCircular(axis_index):
+                    dof_ranges[index] = 2 * numpy.pi
+                else:
+                    dof_ranges[index] = (dof_limit_upper[index]
+                                       - dof_limit_lower[index])
+
+            # Duplicate the logic in or_ompl to convert the DOF resolutions to
+            # a fraction of the longest extent of the state space.
+            maximum_extent = numpy.max(dof_ranges)
```
cdellin: Unfortunately, that's not with the maximum extent is ... you should use `numpy.linalg.norm` instead of `numpy.max`.

src/prpy/planning/ompl.py
```diff
@@ -163,67 +167,115 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
-            ompl_args['range'] = ompl_range
-            logger.debug('Defaulted RRT-Connect range parameter to %.3f.',
-                         ompl_range)
+            conservative_resolution = numpy.min(dof_resolution)
+            conservative_fraction = conservative_resolution / maximum_extent
+
+            if self.multiplier is not None:
+                multiplier = self.multiplier
+            elif self.fraction is not None:
+                multiplier = int(self.fraction / conservative_fraction + 0.5)
```
cdellin: The C side of me likes this, but perhaps we can just use Python's `round` function here?

src/prpy/planning/ompl.py
```diff
@@ -163,67 +167,115 @@ def _Plan(self, robot, goal=None, tsrchains=None, timeout=30., shortcut_timeout=
...
     @ClonedPlanningMethod
     def PlanToConfiguration(self, robot, goal, ompl_args=None, **kw_args):
-        """
-        Plan to a desired configuration with OMPL. This will invoke the OMPL
-        planner specified in the OMPLPlanner constructor.
-        @param robot
-        @param goal desired configuration
-        @return traj
-        """
         ompl_args = self._SetPlannerRange(robot, ompl_args=ompl_args)
```
cdellin: My only other feedback is that `_SetPlannerRange(robot, ompl_args)` is poorly named (it doesn't set the value in the passed `ompl_args` argument, it instead returns a copy with the value set if it doesn't yet exist).  My suggestion would be to rename it to `_CalculatePlannerRange(robot)` which just returns the range float with all dict logic removed, and then replace this line (and the similar line in `PlanToTSR` below) with something like:

```
if ompl_args is None:
    ompl_args = dict()
ompl_args.setdefault('range', _CalculatePlannerRange(robot))
```

mkoval:
@cdellin: I addressed your comments. Could you take another look?

cdellin:
Looks great!

PR #354, opened 2016-08-29
Feature/concatenate trajs

rachelholladay:
Util functionality to concatenate a list of trajectories into one trajectory by appending them onto each other. 

Requested by @gilwoolee

### DIFF
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -2129,3 +2129,38 @@ def GetPointFrom(focus):
         raise ValueError('Focus of the point is an unknown object')
 
     return coord
+
+def concatenateTrajectories(robot, traj_list):
+    """
+    Given a list of trajectories for a single manipulator,
+    concatenate them into a single trajectory
+
+    @param robot
+    @param traj_list List of consective trajectories for a given manipulator
+    @return base_traj Combined trajectory
+    """
+
+    base_traj = traj_list[0]
+    base_cspec = base_traj.GetConfigurationSpecification()
+    traj_indices = base_cspec.ExtractUsedIndices(robot)
+
+    offset = base_traj.GetNumWaypoints()
+    for i in xrange(1, len(traj_list)):
+        traj = traj_list[i]
+        cspec = traj.GetConfigurationSpecification()
+        sub = numpy.subtract(cspec.ExtractUsedIndices(robot), traj_indices)
+        if sub.sum() != 0:
+            raise ValueError('Trajectories are not on the same manipulator.')
+
+        epsilon = 0.01
+        previous_point = base_traj.GetWaypoint(base_traj.GetNumWaypoints()-1)
+        next_point = traj.GetWaypoint(0)
+        #TODO is there a better way to check for this?
+        if numpy.subtract(previous_point, next_point).sum() < epsilon:
+            raise ValueError('Trajectories are not consecutive.')
+
+        for j in xrange(traj.GetNumWaypoints()):
+            waypoint = traj.GetWaypoint(j)
+            base_traj.Insert(offset, waypoint)
+            offset += 1
+    return base_traj
--- a/src/prpy/util.py
+++ b/src/prpy/util.py
@@ -2130,37 +2130,49 @@ def GetPointFrom(focus):
 
     return coord
 
-def concatenateTrajectories(robot, traj_list):
+def concatenateTrajectories(traj_list):
     """
     Given a list of trajectories for a single manipulator,
     concatenate them into a single trajectory
 
-    @param robot
     @param traj_list List of consective trajectories for a given manipulator
     @return base_traj Combined trajectory
     """
+    from planning.base import Tags
+
+    if len(traj_list) == 0:
+        raise ValueError('Trajectory list is empty')
 
     base_traj = traj_list[0]
     base_cspec = base_traj.GetConfigurationSpecification()
-    traj_indices = base_cspec.ExtractUsedIndices(robot)
+    base_robot = base_cspec.ExtractUsedBodies(base_traj.GetEnv())[0]
+    traj_indices = GetTrajectoryIndices(base_traj)
 
     offset = base_traj.GetNumWaypoints()
     for i in xrange(1, len(traj_list)):
-        traj = traj_list[i]
-        cspec = traj.GetConfigurationSpecification()
-        sub = numpy.subtract(cspec.ExtractUsedIndices(robot), traj_indices)
+        traj_i = traj_list[i]
+        cspec_i = traj_i.GetConfigurationSpecification()
+        robot_i = cspec_i.ExtractUsedBodies(traj_i.GetEnv())[0]
+        if robot_i != base_robot:
+            raise ValueError('Trajectories are not on the same robot')
+
+        sub = numpy.subtract(GetTrajectoryIndices(traj_i), traj_indices)
         if sub.sum() != 0:
             raise ValueError('Trajectories are not on the same manipulator.')
 
-        epsilon = 0.01
-        previous_point = base_traj.GetWaypoint(base_traj.GetNumWaypoints()-1)
-        next_point = traj.GetWaypoint(0)
-        #TODO is there a better way to check for this?
-        if numpy.subtract(previous_point, next_point).sum() < epsilon:
-            raise ValueError('Trajectories are not consecutive.')
+        tags_i = GetTrajectoryTags(traj_i)
+        # If one segment is not smooth, the entire trajectory is not (?)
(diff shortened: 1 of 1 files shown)


src/prpy/util.py
```diff
@@ -2129,3 +2129,50 @@ def GetPointFrom(focus):
         raise ValueError('Focus of the point is an unknown object')
 
     return coord
+
+def concatenateTrajectories(traj_list):
```
mkoval: We follow OpenRAVE's naming convention by using `PascalCase` instead of `camelCase`.

src/prpy/util.py
```diff
@@ -2129,3 +2129,50 @@ def GetPointFrom(focus):
...
+    """
+    from planning.base import Tags
+
+    if len(traj_list) == 0:
+        raise ValueError('Trajectory list is empty')
+
+    base_traj = traj_list[0]
+    base_cspec = base_traj.GetConfigurationSpecification()
+    base_robot = base_cspec.ExtractUsedBodies(base_traj.GetEnv())[0]
+    traj_indices = GetTrajectoryIndices(base_traj)
```
mkoval: I suggest strictly handling the case where all of the trajectories have _exactly the same_ `ConfigurationSpecification`. Anything else opens a huge can of worms.

src/prpy/util.py
```diff
@@ -2129,3 +2129,50 @@ def GetPointFrom(focus):
...
+
+        # The entire trajectory gets tagged as constrained if one of 
+        # the segments is constrained.
+        if 'constrained' in tags_i and tags_i['constrained']:
+            SetTrajectoryTags(base_traj, {TAGS.CONSTRAINED: True}, append=True)
+
+        # Add trajectory in by inserting each waypoint
+        for j in xrange(traj_i.GetNumWaypoints()):
+            waypoint = traj_i.GetWaypoint(j)
+            base_traj.Insert(offset, waypoint)
```
mkoval: For example, this would break if the trajectories did not already have the same `ConfigurationSpecification`.

src/prpy/util.py
```diff
@@ -2129,3 +2129,50 @@ def GetPointFrom(focus):
...
+
+        tags_i = GetTrajectoryTags(traj_i)
+        # If one segment is not smooth, the entire trajectory is not (?)
+        if 'smooth' in tags_i and not tags_i['smooth']:
+            SetTrajectoryTags(base_traj, {TAGS.SMOOTH: True}, append=True)
+
+        # The entire trajectory gets tagged as constrained if one of 
+        # the segments is constrained.
+        if 'constrained' in tags_i and tags_i['constrained']:
+            SetTrajectoryTags(base_traj, {TAGS.CONSTRAINED: True}, append=True)
```
mkoval: Replace these repeated calls to `SetTrajectoryTags` with one call at the end. The current logic is incorrect because it will tag a trajectory as `SMOOTH` if *any* of its components are smooth, instead of only when *all* of its components are smooth.

We're also missing a few cases. To be complete, here are all of the tags we should handle:

- `SMOOTH`: `True` if *all* trajectories have this flag set
- `CONSTRAINED`: `True` if *any* trajectory has this flag set
- `DETERMINISTIC_TRAJECTORY`: `True` if *all* trajectories have this flag set
- `DETERMINISTIC_ENDPOINT`: `True` if *the last* trajectory has this flag set

All of the flags should default to `False` if they are not set on a trajectory.

Also, note that setting these tags is a surprisingly expensive operation. It requires parsing XML, parsing JSON, re-serializing JSON, and re-serializing XML.

mkoval:
I posted a few comments. The implementation should actually become quite a bit simpler after applying those changes.

@rachelholladay: Could you take a look?

rachelholladay:
@mkoval I made the changes you requested. The problem I notice (which I didn't before) is that this doesn't check the flags of the first trajectory in the list. The two ways I thought to fix this were

1.  Create a function that checks flags and call it on the first trajectory separately (and in the loop).
2.  Change the loop to iterate through all trajectories, but initialize `base_traj` to be a blank trajectory instead of the first one in the list. 

Would one be preferred over the other? Or some other option I haven't considered?

src/prpy/util.py
```diff
@@ -2129,3 +2129,61 @@ def GetPointFrom(focus):
...
+        # True if any trajectory has this set
+        if 'constrained' in tags_i and tags_i['constrained']:
+            constrainedFlag = True
+
+         # Only True if all trajectories have this set
+        if 'deterministic' in tags_i and not tags_i['deterministic']:
+            deterministcTrajectoryFlag = False
+
+        # True if the last trajectory has this set
+        if i == len(traj_list)-1 and 'deterministic_endpoint' in tags_i and tags_i['determinstic_endpoint']:
```
gilwoolee: Wouldn't it be cleaner to do this outside of the loop?

gilwoolee:
@rachelholladay Thanks for this PR! This concatenation can be further postprocessed to smoothen and shortcut a sequence of trajectories, which would be useful addition. As for the question you asked, I think option 2 (initializing with an empty base trajectory) would be cleaner.

rachelholladay:
@gilwoolee I switched it to using a blank trajectory and even added a unit test. I guess I'd like to have a function argument for whether it should be smoothed or not. What do you think?

src/prpy/util.py
```diff
@@ -2129,3 +2129,75 @@ def GetPointFrom(focus):
...
+    from planning.base import Tags
+
+    if len(traj_list) == 0:
+        raise ValueError('Trajectory list is empty')
+
+    env = robot.GetEnv()
+
+    # Create an empty trajectory to populate
+    base_traj = openravepy.RaveCreateTrajectory(env, '')
+    base_cspec = robot.GetActiveConfigurationSpecification('linear')
```
gilwoolee: Why do we need this to be linear? Also, why are we requiring that the trajectories have to be on the current active dofs? Shouldn't it just be the spec of the first trajectory?

src/prpy/util.py
```diff
@@ -2129,3 +2129,75 @@ def GetPointFrom(focus):
...
+    base_traj.Init(base_cspec)
+    #base_cspec = base_traj.GetConfigurationSpecification()
```
gilwoolee: Please remove this.

src/prpy/util.py
```diff
@@ -2129,3 +2129,75 @@ def GetPointFrom(focus):
...
+    offset = 0
+    for traj in traj_list:
+        cspec_i = traj.GetConfigurationSpecification()
+        if base_cspec != cspec_i:
+            raise ValueError('The configuration specifications of the trajectory do not match')
+
+        tags = GetTrajectoryTags(traj)
+        # Only True if all trajectories have this set
+        if 'smooth' in tags and not tags['smooth']:
+            smoothFlag = False
```
gilwoolee: This would be skipped if `smooth` is not in `tags`. Is `smooth` by default true?

src/prpy/util.py
```diff
@@ -2129,3 +2129,75 @@ def GetPointFrom(focus):
...
+
+        # True if any trajectory has this set
+        if 'constrained' in tags and tags['constrained']:
+            constrainedFlag = True
+
+         # Only True if all trajectories have this set
+        if 'deterministic' in tags and not tags['deterministic']:
```
gilwoolee: Same question.

src/prpy/util.py
```diff
@@ -2129,3 +2129,75 @@ def GetPointFrom(focus):
...
+            deterministcTrajectoryFlag = False
+
+        # Check that trajectory leds from previous,
+        # or if the first, start the trajectory
+        epsilon = 0.001
+        if offset == 0:
+            base_traj.Insert(offset, traj.GetWaypoint(0))
+            offset += 1
+        else:
+            if sum(numpy.subtract(traj.GetWaypoint(0), base_traj.GetWaypoint(offset-1))) > epsilon:
```
gilwoolee: I think we should use `numpy.linalg.norm` here.

tests/test_util.py
```diff
@@ -1088,6 +1088,26 @@ def test_GetMinDistanceBetweenTransformAndWorkspaceTraj(self):
...
+            combined_waypoints[j] = combined_traj.GetWaypoint(j)
+        all_waypoints = [q0, q1, q2]
+
+        error = 'The waypoints on the trajectory do not match the individual waypoints'
+        numpy.testing.assert_array_almost_equal(combined_waypoints,
+                                                all_waypoints,
+                                                decimal=4,
+                                                err_msg=error,
+                                                verbose=True)
+
```
gilwoolee: Can we add tests for checking tags? E.g. when one of the tags along the trajectory is not smooth, etc.

PR #355, opened 2016-08-30
Cleaned up OMPLPlanner

mkoval:
This pull request:

- Cleans up imports in `ompl.py`
- Adds missing docstrings.
- Adds the ability to specify `timelimit` and `ompl_args` in the constructor.
- Replaces `timeout` with `timelimit` to match other PrPy planners.
- Fixes `**kwargs` forwarding in the `OMPLPlanner` wrappers.
- Fixed a bug where `algorithm` was being ignored in favor of the default.


cdellin:
Looks good to me!

PR #356, opened 2016-08-30
Range is actually a distance, not a fraction

cdellin:
This was my mistake -- I maintained yesterday that the `range` parameter is given as a fraction as the maximum extent in the space.  While that's how it's determined by default (via OMPL's [configurePlannerRange](http://ompl.kavrakilab.org/classompl_1_1tools_1_1SelfConfig.html#a65bff53ea4bc6f158342a856175ab9a6)), the parameter itself is actually a distance after all.  Fixing this gives the expected behavior (changing from -epsilon to +epsilon gives a step change in the number of checks needed).


mkoval:
LGTM. I confirmed that `+/-` epsilon changes the number of collision checks.

Thanks @cdellin for getting to the bottom of this.

PR #358, opened 2016-09-08
regrab objects manually in cloned env to set ignore collision list

sjavdani:
When cloning, some properties of grabbed objects isn't being set properly (see discussion in #51 ). This just adds some code to manually regrab objects while setting that information properly. 

Not sure if the RegrabAll() command right above the code I added is still necessary (again, see discussion in #51 ) but it seems like it might be, so I left it there.

### DIFF
--- a/src/prpy/clone.py
+++ b/src/prpy/clone.py
@@ -111,6 +111,13 @@ class Clone(object):
                             cloned_robot = Cloned(robot, into=self.clone_env)
                             cloned_robot.RegrabAll()
 
+                            #regrab manually to set ignore links
+                            for grab_info in robot.GetGrabbedInfo():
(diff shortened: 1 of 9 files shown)


mkoval:
It seems like `RegrabAll` should no longer be necessary. Your implementation looks more correct than `RegrabAll`, which explicitly says that it:

> This has the effect of resetting the current collisions any grabbed body makes with the robot into an ignore list.

@psigen: Do you have a test case that fails without the `RegrabAll()` command? Can you try commenting out the `RegrabAll()` call to test @sjavdani's new implementation?

sjavdani:
Messed up the branch, created a new pull request #364 with new branch

PR #361, opened 2016-11-04
Feature/mimic traj

rachelholladay:
Typically we plan with CBiRRT and we are only concerned with the trajectory of the robot. However, we can perform constrained planning where the robot and an object move together. Our example case has been HERB opening the frig. Therefore this extracts the non-robot trajectories from CBIRRT. To not disrupt the prpy planning pipeline, the class saves the other trajectories such that the user can grab them later. 

There is an unresolved issue that I don't know how to deal with, detailed on line 239.

### DIFF
--- a/src/prpy/planning/cbirrt.py
+++ b/src/prpy/planning/cbirrt.py
@@ -52,6 +52,7 @@ class CBiRRTPlanner(BasePlanner):
         super(CBiRRTPlanner, self).__init__()
 
         self.timelimit = timelimit
+        self.mimic_trajectories = {}
 
         if robot_checker_factory is None:
             robot_checker_factory = DefaultRobotCollisionCheckerFactory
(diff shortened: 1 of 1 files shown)


src/prpy/planning/cbirrt.py
```diff
@@ -234,25 +230,35 @@ def Plan(self, robot, smoothingitrs=None, timelimit=None, allowlimadj=0,
...
                     )
-
+                '''
                 args += ['jointgoals'] + self.serialize_dof_values(goal_config)
 
+                
             if len(jointgoals) > 1:
                 is_endpoint_deterministic = False
 
+        raw_input("Continue?")
```
gilwoolee: This needs to be removed.

src/prpy/planning/cbirrt.py
```diff
@@ -163,7 +164,7 @@ def PlanToTSR(self, robot, tsr_chains, smoothingitrs=100, **kw_args):
 
     def Plan(self, robot, smoothingitrs=None, timelimit=None, allowlimadj=0,
              jointstarts=None, jointgoals=None, psample=None, tsr_chains=None,
-             extra_args=None, **kw_args):
+             extra_args=None, save_mimic_trajectories=False, **kw_args):
         """
         @param allowlimadj If True, adjust the joint limits to include
```
ClintLiddick: Couldn't this also take a dict as an argument and just save the trajectory in there, i.e. a user-supplied out-arg? Then the user could do what they want with the trajectory.

PR #362, opened 2016-11-23
Changes for soft joint limit error handling

Shushman:
This allows the disambiguation between the joint limits from the `openrave Robot` and the actual joint limits of the hardware (from looking up specs). Currently, the two `prpy` planners that call `CheckJointLimits` are now configured to catch `SoftJointLimitError` exceptions and continue.

@gilwoolee please take a look. There aren't many changes, and this is definitely still in progress. We also need to something about OMPL planners.

### DIFF
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -544,7 +544,7 @@ class HardwareJointLimitValidator(MetaPlanner):
             try:
                 result =  plan_fn(*args, **kw_args)
             except UnsupportedPlanningError:
-                continue
+                pass
             finally:
                 if resetLimits == True:
                     robot.SetDOFLimits(soft_limit_min, soft_limit_max)
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -466,6 +466,96 @@ class MethodMask(MetaPlanner):
         else:
             raise UnsupportedPlanningError()
 
+
+class HardwareJointLimitValidator(MetaPlanner):
+    def __init__(self, planner):
+        super(HardwareJointLimitValidator, self).__init__()
+        self._planner = planner
+
+    def __str__(self):
+        return 'HardLimitValidator({:s})'.format(', '.join(map(str, self._planners)))
+
+    def get_planners(self, method_name):
+        if planner.has_planning_method(method_name):
+            return [self._planner]
+        else:
+            return []
+
+    def plan(self, method, args, kw_args):
+
+        # TODO : Is this correct?
+        robot = args[0]
+
+        #Get HARD DOF Limits from robot
+        #hard_limit_min, hard_limit_max = robot.GetHardwareDOFLimits()
+        active_dof_indices = robot.GetActiveDOFIndices()
+        currentDOFVals = robot.getActiveDOFValues()
+        soft_limit_min, soft_limit_max = robot.GetActiveDOFLimits()
+        resetLimits = False
+
+        lower_hard_violations = (currentDOFVals < hard_limit_min)
+        upper_hard_violations = (currentDOFVals > hard_limit_max)
+        lower_soft_violations = (currentDOFVals < soft_limit_min)
+        upper_soft_violations = (currentDOFVals > soft_limit_max)
+
+        if lower_soft_violations.any() || upper_soft_violations.any():
+            
+            # Check if hard limits violated
+            if lower_hard_violations.any():
+                index = lower_hard_violations.nonzero()[0][0]
+                raise HardJointLimitError(
+                    robot,
+                    dof_index=active_dof_indices[index],
+                    dof_value=currentDOFVals[index],
+                    dof_limit=hard_limit_min[index],
+                    description='position',
+                    deterministic=deterministic)
+            if upper_hard_violations.any():
+                index = upper_hard_violations.nonzero()[0][0]
+                raise HardJointLimitError(
+                    robot,
+                    dof_index=active_dof_indices[index],
+                    dof_value=currentDOFVals[index],
+                    dof_limit=hard_limit_max[index],
+                    description='position',
+                    deterministic=deterministic)
+
+            # Between software and hardware limits
+            # Temporarily extend joint limits as needed
+            soft_limit_min_new = soft_limit_min
+            soft_limit_max_new = soft_limit_max
+
+            if lower_soft_violations.any():
+                soft_limit_min_new[lower_soft_violations] = currentDOFVals[lower_soft_violations]
+            if upper_soft_violations.any():
+                soft_limit_max_new[upper_soft_violations] = currentDOFVals[upper_soft_violations]
+
+            robot.SetDOFLimits(soft_limit_min_new, soft_limit_max_new)
+
+            resetLimits = True
+
+        result = None    
+
+        if planner.has_planning_method(method):
+            plan_fn = getattr(planner, method)
+
+            try:
+                result =  plan_fn(*args, **kw_args)
+            except UnsupportedPlanningError:
+                continue
+
+        # Reset limits if they were changed
+        if resetLimits == True:
+            robot.SetDOFLimits(soft_limit_min, soft_limit_max)
+
+        # Return result of planner if applicable
+        if result is not None:
+            return result
+        else:
+            raise UnsupportedPlanningError()
+
+                
+
 @contextmanager
 def save_dof_limits(robot):
     upper, lower = robot.GetDOFLimits()
--- a/src/prpy/planning/base.py
+++ b/src/prpy/planning/base.py
@@ -468,9 +468,10 @@ class MethodMask(MetaPlanner):
 
 
 class HardwareJointLimitValidator(MetaPlanner):
-    def __init__(self, planner):
+    def __init__(self, planner,hard_limits):
         super(HardwareJointLimitValidator, self).__init__()
         self._planner = planner
+        self.hard_limits = hard_limits
 
     def __str__(self):
         return 'HardLimitValidator({:s})'.format(', '.join(map(str, self._planners)))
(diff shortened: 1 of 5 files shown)


src/prpy/util.py
```diff
@@ -1359,7 +1361,33 @@ def CheckJointLimits(robot, q, deterministic=None):
...
+        raise SoftJointLimitError(
+            robot,
+            dof_index=active_dof_indices[index],
+            dof_value=q[index],
+            dof_limit=q_limit_max[index],
+            description='position',
+            deterministic=deterministic)
+
+
+    # Soft Joint Limit Error implies Hard Limit Error but not vice versa
```
mkoval: We should verify somewhere (maybe in `Robot`?) that this is the case.

src/prpy/util.py
```diff
@@ -1326,7 +1326,7 @@ def ComputeGeodesicUnitTiming(traj, env=None, alpha=1.0):
     return new_traj
 
 
-def CheckJointLimits(robot, q, deterministic=None):
+def CheckJointLimits(robot, q, hard_limits, deterministic=None):
```
psigen: This is a little weird because the only reason we store the soft limits is because we are using OpenRAVE's `GetActiveDOFLimits()` function, which isn't integral to `prpy`.  So adding the hard limits is not quite analogous, even if we put it in the `prpy.base.Robot`.

Can someone bring me up to speed on what the usage pattern for soft vs. hard limits is intended to be?

gilwoolee: @psigen My understanding is that we only use soft limit during planning so when the robot starts at a hard joint limit the planner would through a JointLimitError, but we'd like this to simply plan through instead of erroring out.

cdellin-nuro: This type of logic does seem robot-specific though -- perhaps it would make more sense to put this sort of logic in `herbpy` instead, in the `Planner` wrapper? The wrapper could simply extend the joint limits temporarily (up to the hard limits) before invoking the planner. That way each planner would still only have to know about one set of limits.

src/prpy/util.py
Shushman: Well actually I think I typo'd there. I believe the prevailing logic was the other way around, i.e. the joint limits from `OpenRAVE` are _more_ restrictive than the actual joint limits?

src/prpy/util.py
Shushman: Well the hard limits are indeed robot specific, which is why I thought to send them as an argument, but can we just make them a property of the robot then?
As for the planner's awareness to limits, I figured to use the `ompl` planners for example, where `or_ompl` enforces the joint limits, we would anyway have to change that logic to accommodate both kinds of limits right? Since `or_ompl` planners do not interact with `prpy.util`

cdellin-nuro: I was thinking that the `Planner` wrapper in `herbpy` would:

1. Check if the starting configuration is outside of `robot.GetDOFLimits()`. If it's not, then just continue as normal.
1. Otherwise, check if the starting configuration is outside of the hardware limits. If so, then there's some sort of error (perhaps the offsets are all wrong?) and the planner should definitely throw an error.
1. If we get here, then one or more of the starting configuration DOF values are outside the `DOFLimits` but inside the hardware limits. Compute new limits by extending the `DOFLimits` just enough to include the starting configuration, and set them on the robot temporarily by calling `SetDOFLimits()`. Then call the underlying planner as usual.  Make sure to reset the `DOFLimits` when the planner finishes (even in cases where the planner throws an exception) using a `finally` statement or a state saver.

The advantage here is that each planner (e.g. `or_ompl`) can continue to just use `GetDOFLimits()` and don't need to change. This logic could even just be implemented as a `prpy` meta-planner, which could be inserted into the `herbpy` planning pipeline.

cdellin: This type of logic does seem robot-specific though -- perhaps it would make more sense to put this sort of logic in `herbpy` instead, in the `Planner` wrapper? The wrapper could simply extend the joint limits temporarily (up to the hard limits) before invoking the planner. That way each planner would still only have to know about one set of limits.

I was thinking that the `Planner` wrapper in `herbpy` would:

1. Check if the starting configuration is outside of `robot.GetDOFLimits()`. If it's not, then just continue as normal.
1. Otherwise, check if the starting configuration is outside of the hardware limits. If so, then there's some sort of error (perhaps the offsets are all wrong?) and the planner should definitely throw an error.
1. If we get here, then one or more of the starting configuration DOF values are outside the `DOFLimits` but inside the hardware limits. Compute new limits by extending the `DOFLimits` just enough to include the starting configuration, and set them on the robot temporarily by calling `SetDOFLimits()`. Then call the underlying planner as usual.  Make sure to reset the `DOFLimits` when the planner finishes (even in cases where the planner throws an exception) using a `finally` statement or a state saver.

The advantage here is that each planner (e.g. `or_ompl`) can continue to just use `GetDOFLimits()` and don't need to change. This logic could even just be implemented as a `prpy` meta-planner, which could be inserted into the `herbpy` planning pipeline.

mkoval: I like @cdellin's suggestion quite a bit. It makes a lot of sense to implement this as a `MetaPlanner` that temporarily extends the robot's joint limits. It's also quite a bit simpler to implement than this pull request, since it only need be implemented once - instead of once per planner.

Shushman: Thanks! @mkoval @cdellin I made a first pass at a Meta Planner in `...planning/base.py`. Can you glance over it and tell me if it seems to be doing the right thing? I'll refine accordingly.

src/prpy/planning/base.py
```diff
@@ -466,6 +466,96 @@ def plan(self, method, args, kw_args):
...
+        else:
+            return []
+
+    def plan(self, method, args, kw_args):
+
+        # TODO : Is this correct?
+        robot = args[0]
+
+        #Get HARD DOF Limits from robot
+        #hard_limit_min, hard_limit_max = robot.GetHardwareDOFLimits()
```
gilwoolee: Don't we need `prpy.base.robot` to implement `GetHardwareDOFLImits()`?

src/prpy/planning/base.py
```diff
@@ -466,6 +466,96 @@ def plan(self, method, args, kw_args):
...
+        active_dof_indices = robot.GetActiveDOFIndices()
+        currentDOFVals = robot.getActiveDOFValues()
+        soft_limit_min, soft_limit_max = robot.GetActiveDOFLimits()
+        resetLimits = False
+
+        lower_hard_violations = (currentDOFVals < hard_limit_min)
+        upper_hard_violations = (currentDOFVals > hard_limit_max)
+        lower_soft_violations = (currentDOFVals < soft_limit_min)
+        upper_soft_violations = (currentDOFVals > soft_limit_max)
```
gilwoolee: I'm assuming that `robot.GetHardwareDOFLimits` will return `(-inf, inf)` if there is no limit. Otherwise, this will need to handle some special cases.

src/prpy/planning/base.py
```diff
@@ -466,6 +466,96 @@ def plan(self, method, args, kw_args):
...
+        result = None    
+
+        if planner.has_planning_method(method):
+            plan_fn = getattr(planner, method)
+
+            try:
+                result =  plan_fn(*args, **kw_args)
+            except UnsupportedPlanningError:
+                continue
+
```
gilwoolee: Add a `finally` here and call `SetDOFLimits` if `resetLimits = True`. Otherwise, `plan_fn` might throw an uncaught error and this might return without setting the DOFs back.

src/prpy/util.py
gilwoolee: Revert these changes.

src/prpy/planning/snap.py
```diff
@@ -106,7 +110,11 @@ def _Snap(self, robot, goal, **kw_args):
         # Make the trajectory end at the goal configuration, as
         # long as it is not in collision and is not identical to
         # the start configuration.
-        CheckJointLimits(robot, goal, deterministic=True)
+        try:
+            CheckJointLimits(robot, goal, deterministic=True)
+        except SoftJointLimitError:
+            pass
```
gilwoolee: Wouldn't it be better if `CheckJointLimits` returns `SoftJointLimitError`?

src/prpy/planning/base.py
cdellin: One idea would be to just pass the hard limits into the `HardwareJointLimitValidator` instance on construction. Then you don't need to add any additional robot APIs.

Shushman: I like the idea of @cdellin . Not just because it means less work.

src/prpy/util.py
Shushman: Is there actually a need for `CheckJointLimits` now? Since every planning call wraps the `HardwareJointLimitValidator` in it?

src/prpy/planning/base.py
```diff
@@ -466,6 +466,96 @@ def plan(self, method, args, kw_args):
...
+        currentDOFVals = robot.getActiveDOFValues()
+        soft_limit_min, soft_limit_max = robot.GetActiveDOFLimits()
+        resetLimits = False
+
+        lower_hard_violations = (currentDOFVals < hard_limit_min)
+        upper_hard_violations = (currentDOFVals > hard_limit_max)
+        lower_soft_violations = (currentDOFVals < soft_limit_min)
+        upper_soft_violations = (currentDOFVals > soft_limit_max)
+
+        if lower_soft_violations.any() || upper_soft_violations.any():
```
Shushman: Done. Thanks, my MATLAB-ing is getting in my head now.

src/prpy/planning/base.py
Shushman: Done.

src/prpy/util.py
Shushman: I've reverted the changes but my original question about needing `CheckJointLimits` remains. Thoughts? @cdellin @mkoval ?

src/prpy/planning/snap.py
Shushman: Well doesn't `raise` mean throwing the error? And `CheckJointLimits` does raise an error, we were just ignoring it. This may change with the new API suggested by Chris though. I've reverted `CheckJointLimits` to just return a `SoftJointLimitError`. But it seems that this is no longer necessary anyway since the `HardwareJointLimitValidator` will take care of that case?

src/prpy/util.py
Shushman: Thus far this PR only adds potential functionality through the `HardwareJointLimitValidator` class but does not change the logic of anything else.
The issue is that changing the logic as suggested here will cause some `prpy` unit tests to fail. A unit test for checking invalid limits expects an error to be thrown, but under this new API it will just temporarily extend the joint limits rather than throw an error. Should we edit those unit tests @mkoval ?

Shushman:
Closing since `prpy` is no longer supported. this is being introduced in libherb

PR #363, opened 2016-12-05
Generate documentation with Doxygen

brianhou:


brianhou:
There were some broken links in the README. I fixed two, but the vncc repo seems to no longer exist so I don't know what to link to there. Trollius is deprecated, but I updated the link anyway.

gilwoolee:
@brianhou `vncc` and `vncc_msgs` repos exist, but they are private. I think we should leave it in to make it explicit for us, but I'm not sure since prpy is meant to be public. @ClintLiddick what do you think?

ClintLiddick:
This looks good to me. Thanks @brianhou !

PR #364, opened 2016-12-09
regrab objects manually in clone

sjavdani:
(I messed up the branch for the other pull request, #358, seemed cleanest just to delete that and start over)

When cloning, some properties of grabbed objects isn't being set properly (see discussion in #51 ). This just adds some code to manually regrab objects while setting that information properly.

Note that the main ada demo doesn't work without this change, as the fork is continuously in collision with the fingers.

### DIFF
--- a/src/prpy/clone.py
+++ b/src/prpy/clone.py
@@ -109,7 +109,14 @@ class Clone(object):
                                     .format(robot.GetName())
                                 )
                             cloned_robot = Cloned(robot, into=self.clone_env)
-                            cloned_robot.RegrabAll()
+
+                            #regrab manually to set ignore links
+                            for grab_info in robot.GetGrabbedInfo():
+                                item_to_grab = self.clone_env.GetKinBody(grab_info._grabbedname)
+                                grablink = cloned_robot.GetLink(grab_info._robotlinkname)
+                                linkstoignore = grab_info._setRobotLinksToIgnore
+                                cloned_robot.Grab(item_to_grab, grablink=grablink, linkstoignore=linkstoignore)
+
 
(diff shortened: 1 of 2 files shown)


sjavdani:
We're starting to share ada code with collaborators - can we get this fixed before that, so we don't need to manage a different branch to make the ada demo work?

psigen:
@mkoval @cliddick I think this is probably fine, right? Do we already have
unit tests for grabbing?

On Dec 9, 2016 17:16, "Shervin Javdani" <notifications@github.com> wrote:

We're starting to share ada code with collaborators - can we get this fixed
before that, so we don't need to manage a different branch to make the ada
demo work?

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/personalrobotics/prpy/pull/364#issuecomment-266138240>,
or mute the thread

gilwoolee:
We don't have unit tests for grabbing. @sjavdani Would you mind creating a grabbing test that tests this issue?

sjavdani:
@gilwoolee sure. Can you point me to a similar set of tests so I have a place to start?

Also, how can I go about doing this in prpy, without using herbpy and adapy?

sjavdani:
Ok, I added unit tests for grasping and cloning, and making sure attributes don't change.

However, I wasn't actually able to create the problem I saw with ada on the wam...so, I'm not sure what this confirms, other than my change didn't break what wasn't problematic with the wam

gilwoolee:
@sjavdani Thanks for the changes and the tests. Your changes seem to be fine, but if you see the same problem again, we should create a test case.

PR #365, opened 2016-12-13
Fix two bugs in Future

mkoval:
- Fix a logic error in `add_done_callback`.
- Fix reference to non-existent logger in `_set_done`.

### DIFF
--- a/src/prpy/futures.py
+++ b/src/prpy/futures.py
@@ -167,7 +167,7 @@ class Future(object):
         @type  fn: (ResultType) -> None
         """
         with self.lock:
-            if self._is_done:
+            if not self._is_done:
                 if fn in self._callbacks:
                     raise ValueError('Callback is already registered.')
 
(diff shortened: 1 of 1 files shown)


src/prpy/futures.py
```diff
@@ -167,7 +167,7 @@ def add_done_callback(self, fn):
         @type  fn: (ResultType) -> None
         """
         with self.lock:
-            if self._is_done:
+            if not self._is_done:
                 if fn in self._callbacks:
```
psigen: Should this check be moved outside this `if` block entirely? it seems like it applies to both cases.

cdellin:
Wow, you guys are already fixing bugs in the future? All the bugs in the present are already solved?

gilwoolee:
Haha yeah that's what I thought too when I first got this email.

- Gilwoo

src/prpy/futures.py
mkoval: It does seem that way. 👍

gilwoolee:
Thanks @mkoval !

PR #367, opened 2017-02-24
Split tsr

Shushman:
This branch removes the `tsr` subfolder from the prpy package and changes the imports to reflect that.


brianhou:
Tests fail due to an assortment of ImportErrors. The root cause is a `tsr.py` in `prpy.planning` that is found before the `tsr` module. Renaming `tsr.py` seems to fix that issue. I'll let you pick the new name.

In addition, there were a couple of places that passed the `manip` parameter to `TSR`, which seems like it should be `manipindex`. Changing that in 8dbb20adc8b746562e27b481b51d8c389157df88 made the tests pass.

Shushman:
These are the changes so that `prpy/tests/planning` now all run:

- Renamed `tsr.py` to `tsrplanner.py`
- Changed `__init__.py` to import from `tsrplanner`

Shushman:
The final change is to `base/robot.py` to change the constructor call to `tsrlibrary`
The call now provides the `manipindex` argument to be compatible with ` tsr` without OpenRAVE dependencies

PR #371, opened 2017-04-03
Editing init.py for backwards compatibility

Shushman:
This is a temporary fix while we figure out something permanent so that an `import prpy.tsr.XYZ` statement still works.


src/prpy/tsr.py
```diff
@@ -0,0 +1,34 @@
...
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+import warnings
+
+warnings.warn(
```
brianhou: This should be a syntax error?

Shushman: Damn I forgot to run checkout on this. This won't be there for now.

Shushman: I'll close this PR and make a separate one on a fresh branch to keep a cleaner commit history

PR #372, opened 2017-04-03
Edit __init__.py for backwards compatibility with TSR

Shushman:
This avoids breaking changes for code that imports `prpy.tsr`


Shushman:
Having added the stub directory for `tsr` under `prpy/src/prpy/`, any imports in dependent packages of the form `import prpy.tsr.XYZ` will work.

mkoval:
Thanks @Shushman for making this change! 🎉

PR #377, opened 2021-01-23
Fix Python syntax error in meta.py

rsinnet:
Class definitions didn't have colons. This would cause an error if trying to import `meta.py`. This will cause an error in installation as well in the step where `.py` files are compiled to `.pyc` files, which is where I noticed it.


brianhou (approved):
Thanks for the fix!

